PR #28, opened 2023-06-09
Update sqlc to v1.18.0

bplunkett-stripe:
Upgrade sqlc to v1.18.0.

All of the `name` columns (really an alias for varchar) sqlc generated code used the go type `interface{}`. Manually casting it to `::TEXT` fixed the code gen. I opted to use `::TEXT` rather than `::VARCHAR` because it's slightly more flexible 🤷‍♂️


internal/queries/queries.sql
```diff
@@ -37,18 +37,19 @@ WHERE a.attrelid = $1
 ORDER BY a.attnum;
 
 -- name: GetIndexes :many
-SELECT c.oid                                   AS oid,
-       c.relname                               as index_name,
-       (SELECT c1.relname AS table_name FROM pg_catalog.pg_class c1 WHERE c1.oid = i.indrelid),
```
bplunkett-stripe, line 42: Clean up this query a bit

Issue #34, opened 2023-06-26
Sequence (and Serial) support


bplunkett-stripe:
Add support for serials (and sequences).

We initially won't support altering sequences, but this should unblock some folks, since serials are frequently used.

bplunkett-stripe:
Sequence support added in PR above

PR #35, opened 2023-06-28
Add sequence support

bplunkett-stripe:
### Description
Add sequence (and serial support). A little gross because we all columns ops are grouped in the same add/alter block.

### Motivation
https://github.com/stripe/pg-schema-diff/issues/34

### Testing
Tested via acceptance tests

### DIFF
--- a/internal/migration_acceptance_tests/sequence_cases_test.go
+++ b/internal/migration_acceptance_tests/sequence_cases_test.go
@@ -69,6 +69,7 @@ var sequenceAcceptanceTests = []acceptanceTestCase{
 			`,
 		},
 		expectedHazardTypes: []diff.MigrationHazardType{
+			diff.MigrationHazardTypeDeletesData,
 			diff.MigrationHazardTypeHasUntrackableDependencies,
 		},
 	},
@@ -136,6 +137,7 @@ var sequenceAcceptanceTests = []acceptanceTestCase{
 		},
 		expectedHazardTypes: []diff.MigrationHazardType{
 			diff.MigrationHazardTypeAcquiresAccessExclusiveLock,
+			diff.MigrationHazardTypeDeletesData,
 			diff.MigrationHazardTypeImpactsDatabasePerformance,
 		},
 	},
@@ -165,6 +167,7 @@ var sequenceAcceptanceTests = []acceptanceTestCase{
 		},
 		expectedHazardTypes: []diff.MigrationHazardType{
 			diff.MigrationHazardTypeAcquiresAccessExclusiveLock,
+			diff.MigrationHazardTypeDeletesData,
 			diff.MigrationHazardTypeImpactsDatabasePerformance,
 		},
 	},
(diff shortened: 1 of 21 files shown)


pkg/diff/sql_generator.go
```diff
@@ -49,19 +49,23 @@ var (
...
 }
 
 func (o oldAndNew[S]) GetNew() S {
-	return o.old
+	return o.new
 }
 
 func (o oldAndNew[S]) GetOld() S {
-	return o.new
+	return o.old
```
bplunkett-stripe, line 68: Crazy bug. I'm shocked this was working, but I'm guessing it's because we only use this functions in a few places

pkg/diff/sql_generator.go
```diff
@@ -1228,6 +1288,176 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+	sb.WriteString(fmt.Sprintf("\tINCREMENT BY %d\n", seq.Increment))
+	sb.WriteString(fmt.Sprintf("\tMINVALUE %d MAXVALUE %d\n", seq.MinValue, seq.MaxValue))
+	cycleModifier := ""
+	if !seq.Cycle {
+		cycleModifier = "NO "
+	}
+	sb.WriteString(fmt.Sprintf("\tSTART WITH %d CACHE %d %sCYCLE\n", seq.StartValue, seq.CacheSize, cycleModifier))
+
+	var hazards []MigrationHazard
+	if seq.Owner == nil {
```
alexaub-stripe, line 1309: Should there be an OWNED BY statement in the create sequence if there is an owner?

pkg/diff/sql_generator.go
```diff
@@ -1228,6 +1288,176 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+	}}, nil
+}
+
+func (s *sequenceSQLVertexGenerator) Alter(diff sequenceDiff) ([]Statement, error) {
+	// Ownership changes handled by the sequenceOwnershipSQLVertexGenerator
+	diff.old.Owner = diff.new.Owner
+	if !cmp.Equal(diff.old, diff.new) {
+		// Technically, we could support altering expression, but I don't see the use case for it. it would require more test
+		// cases than forceReadding it, and I'm not convinced it unlocks any functionality
+		return nil, fmt.Errorf("altering check constraint to resolve the following diff %s: %w", cmp.Diff(diff.old, diff.new), ErrNotImplemented)
```
alexaub-stripe, line 1340: This error appears to refer to check constraints and should refer to sequence alters.

internal/migration_acceptance_tests/sequence_cases_test.go
```diff
@@ -0,0 +1,441 @@
...
+
+			ALTER SEQUENCE "foobar sequence" OWNED BY some_other_foobar."some id";
+			`,
+		},
+		expectedHazardTypes: []diff.MigrationHazardType{
+			diff.MigrationHazardTypeDeletesData,
+		},
+	},
+	{
+		name: "Alter ownership data type",
```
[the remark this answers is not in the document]
alexaub-stripe: Ah i see we're not supporting alters today. Maybe we should test that we get errors for unsupported operations? Or do we always just recreate?

pkg/diff/sql_generator.go
```diff
@@ -1228,6 +1288,176 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+		Timeout: statementTimeoutDefault,
+		Hazards: hazards,
+	}}, nil
+}
+
+func (s *sequenceSQLVertexGenerator) Alter(diff sequenceDiff) ([]Statement, error) {
+	// Ownership changes handled by the sequenceOwnershipSQLVertexGenerator
+	diff.old.Owner = diff.new.Owner
+	if !cmp.Equal(diff.old, diff.new) {
+		// Technically, we could support altering expression, but I don't see the use case for it. it would require more test
```
alexaub-stripe, line 1338: I could imagine for example wanting to increase the max value without resetting the current value, but I'm largely fine with this sentiment in a declarative schema world. We can wait an see desire for this.

pkg/diff/sql_generator.go
```diff
@@ -1228,6 +1288,176 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+	var hazards []MigrationHazard
+	if seq.Owner != nil && (s.isDeletedWithOwningTable(seq) || s.isDeletedWithColumns(seq)) {
+		return nil, nil
+	} else if seq.Owner == nil {
+		hazards = append(hazards, migrationHazardSequenceCannotTrackDependencies)
+	}
+	return []Statement{{
+		DDL:     fmt.Sprintf("DROP SEQUENCE %s", seq.GetFQEscapedName()),
+		Timeout: statementTimeoutDefault,
+		Hazards: hazards,
```
alexaub-stripe, line 1330: Dropping a sequence will destroy state (the internal value), so there should be a hazard for that right? This should trigger on recreates in the absence of alters.

pkg/diff/sql_generator.go
```diff
@@ -1228,6 +1288,176 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+}
+
+func (s *sequenceSQLVertexGenerator) GetDeleteDependencies(seq schema.Sequence) []dependency {
+	var deps []dependency
+	// This is an unfortunate hackaround. It would make sense to also have a dependency on the owner column, such that
+	// the sequence can only be considered deleted after the owning column is deleted. However, we currently don't separate
+	// column deletes from table add/alters. We can't build this dependency without possibly creating a circular dependency:
+	// the sequence add/alter must occur before the new owner column add/alter, but the sequence delete must occur after the
+	// old owner column delete (equivalent to add/alter) and the sequence add/alter. We can get away with this because
+	// we, so far, no columns are ever "re-created". If we ever do support that, we'll need to revisit this.
```
alexaub-stripe, line 1368:
```suggestion
	// old owner column delete (equivalent to add/alter) and the sequence add/alter. We can get away with this because
	// so far, no columns are ever "re-created". If we ever do support that, we'll need to revisit this.
```

pkg/diff/sql_generator.go
alexaub-stripe: Ah is this handled by the sequenceOwnershipSQLVertexGenerator?

alexaub-stripe (changes requested):
Looks great! Comments are worth looking at so requesting changes, but this largely lgtm

pkg/diff/sql_generator.go
bplunkett-stripe: Yep! It's split up because the ownership statement rarely comes directly after the sequence being created. Serials:
```
CREATE SEQUENCE
CREATE TABLE
ASSIGN OWNERSHIP OF SEQUENCE TO TABLE
```

internal/queries/queries.sql
```diff
@@ -90,16 +90,25 @@ FROM pg_catalog.pg_proc proc
 WHERE proc_namespace.nspname = 'public'
   AND proc.prokind = 'f'
   -- Exclude functions belonging to extensions
-  AND NOT EXISTS(SELECT depend.objid FROM pg_catalog.pg_depend depend WHERE deptype = 'e' AND depend.objid = proc.oid);
+  AND NOT EXISTS(
+        SELECT depend.objid
+        FROM pg_catalog.pg_depend depend
+        WHERE depend.classid = 'pg_proc'::regclass
+          AND depend.objid = proc.oid
```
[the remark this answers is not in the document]
bplunkett-stripe: I think it's better to be safe. This  seems to be the norm, so I'm going to stick with it.

internal/migration_acceptance_tests/sequence_cases_test.go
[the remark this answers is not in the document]
bplunkett-stripe: We error out. I can add a test case for each

internal/schema/schema.go
```diff
@@ -220,6 +215,29 @@ func (c CheckConstraint) GetName() string {
 	return c.Name
 }
 
+type (
+	// SequenceOwner represents the owner of a sequence. Once we remove TableUnescapedName, we can replace it with
+	// ColumnIdentifier struct that can also be used in the Column struct
+	SequenceOwner struct {
+		TableName SchemaQualifiedName
+		// TableUnescapedName is a hackaround until we switch over Tables ot use fully-qualified, escaped names
+		TableUnescapedName string
```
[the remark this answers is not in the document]
bplunkett-stripe: The issue stems from using the unescaped name in the `Table` schema object. I don't think marking the field as deprecated indicates where the problem actually arises from (the `Table` schema object's logic for generating name). I do have a ticket to fix this though!

pkg/diff/sql_generator.go
bplunkett-stripe: Good spot!

pkg/diff/sql_generator.go
bplunkett-stripe: Yep, we can add support for this as folks ask for it. Should be pretty straightforward

pkg/diff/sql_generator.go
```diff
@@ -1318,7 +1318,10 @@ func (s *sequenceSQLVertexGenerator) Add(seq schema.Sequence) ([]Statement, erro
 }
 
 func (s *sequenceSQLVertexGenerator) Delete(seq schema.Sequence) ([]Statement, error) {
-	var hazards []MigrationHazard
+	hazards := []MigrationHazard{{
+		Type:    MigrationHazardTypeDeletesData,
+		Message: "By deleting a sequence, its state will be permanently lost",
```
alexaub-stripe, line 1323: nit: Maybe replace "its state" with "its current value" to be specific?

alexaub-stripe (approved):

alexaub-stripe (approved):

PR #44, opened 2023-07-14
SQL linting

bplunkett-stripe:
### Description
Uses SQLfluff for linting. This linter requires folks to have python installed or use docker. There aren't any super great-looking go SQL linters, and most folks have Python installed on their machines, so it should be fine.

### Motivation
Enforce a consistent style in SQL

### Testing
Tested locally and in CI

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -15,7 +15,8 @@ jobs:
         run: docker build -t pg-schema-diff-lint-runner -f ./build/Dockerfile.lint .
       - name: Run lint
         run: docker run pg-schema-diff-lint-runner
-  # The "lint" job will lint go (and SQL), but the "go_lint" job will also highlight specific go errors in the PR
+  # The "lint" job will lint go (and SQL). The "go_lint" job will also lint go but adds comments to the PR on the lines
+  # containing the linting errors
   go_lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -1,19 +1,29 @@
-.PHONY: code_gen format go_mod_tidy lint test vendor sqlc
+.PHONY: code_gen format go_lint go_mod_tidy lint sqlc sql_lint vendor
 
-format:
-	goimports -w .
+code_gen: go_mod_tidy sqlc
 
-lint:
+go_lint:
```
bplunkett-stripe, line 5: Alphabetically organized make targets

.github/workflows/lint.yaml
```diff
@@ -8,6 +8,15 @@ permissions:
   contents: read
 jobs:
   lint:
+    runs-on: ubuntu-latest
```
bplunkett-stripe, line 11: We currently have two jobs that run go linting. Why two?

One job runs linting for all languages (go, sql) and any languages we might add in the future. 

The go lint job also lints go (using the same rules), but will add comments to the PR where the linting violations are.

Reason for both jobs linting go:
If someone is running linting locally, they can just run the lint docker container without needing to install golangci-lint. I feel making the golinter togglable in the container is unnecessarily right now (seems fine that the same linting rules are being run twice)

bplunkett-stripe: image

Example attached

alexaub-stripe (approved):

PR #48, opened 2023-07-24
chore: simplify range expression

okazaki-kk:
### Description
Sorry for the very small change, but I got a format warning on vscode. So I simplified the variable of range loop.

### Motivation

### Testing

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -87,7 +87,7 @@ func (g *Graph[V]) HasVertexWithId(id string) bool {
 // Reverse reverses the edges of the map. The sources become the sinks and vice versa.
 func (g *Graph[V]) Reverse() {
 	reversedEdges := make(AdjacencyMatrix)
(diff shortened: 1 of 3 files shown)


bplunkett-stripe:
Hi! 

We appreciate taking the time to improve the codebase! I think you make a good point here; however, I believe if this is something we want to enforce, it should be enforced through a linter and not by-hand.If this is blocking you from development, you could try disabling the linting rules on your own editor.

The reason I structured loops like this (feel free to disagree), 
`for k, _ := range myMap` is I feel it's more explicit when iterating over k,v pairs. I feel like there's some ambiguity when doing `for k := range myMap`, i.e., is it iterating over the keys or vals. 

I do think it's more idiomatic to do it `for k := range myMap`, so I think investing in the linting rule might be the best course of action rather than spot fixing. Definitely let me know if you have any thoughts/questions, and feel free to re-open the issue!

PR #53, opened 2023-08-07
Add foreign key support

bplunkett-stripe:
### Description
Add foreign key support

### Motivation
https://github.com/stripe/pg-schema-diff/issues/49

### Testing
Tested with acceptance tests

### DIFF
--- a/internal/queries/queries.sql
+++ b/internal/queries/queries.sql
@@ -68,7 +68,9 @@ SELECT
 FROM pg_catalog.pg_class AS c
 INNER JOIN pg_catalog.pg_index AS i ON (i.indexrelid = c.oid)
 INNER JOIN pg_catalog.pg_class AS table_c ON (table_c.oid = i.indrelid)
-LEFT JOIN pg_catalog.pg_constraint AS con ON (con.conindid = c.oid)
+LEFT JOIN
+    pg_catalog.pg_constraint AS con
+    ON (con.conindid = c.oid AND con.contype IN ('p', 'u', NULL))
 LEFT JOIN
     pg_catalog.pg_inherits AS idx_inherits
     ON (c.oid = idx_inherits.inhrelid)
@@ -107,6 +109,35 @@ WHERE
     pg_class.relnamespace
     = (SELECT oid FROM pg_catalog.pg_namespace WHERE nspname = 'public')
     AND pg_constraint.contype = 'c'
+    AND pg_constraint.conislocal;
+
+-- name: GetForeignKeyConstraints :many
+SELECT
+    pg_constraint.conname::TEXT AS constraint_name,
+    constraint_c.relname::TEXT AS owning_table_name,
+    constraint_namespace.nspname::TEXT AS owning_table_schema_name,
+    foreign_table_c.relname::TEXT AS foreign_table_name,
+    foreign_table_namespace.nspname::TEXT AS foreign_table_schema_name,
+    pg_constraint.convalidated AS is_valid,
+    pg_catalog.pg_get_constraintdef(pg_constraint.oid) AS constraint_def
+FROM pg_catalog.pg_constraint
+INNER JOIN
+    pg_catalog.pg_class AS constraint_c
+    ON pg_constraint.conrelid = constraint_c.oid
+INNER JOIN pg_catalog.pg_namespace AS constraint_namespace
+    ON pg_constraint.connamespace = constraint_namespace.oid
+INNER JOIN
+    pg_catalog.pg_class AS foreign_table_c
+    ON pg_constraint.confrelid = foreign_table_c.oid
+INNER JOIN pg_catalog.pg_namespace AS foreign_table_namespace
+    ON
+        foreign_table_c.relnamespace = foreign_table_namespace.oid
+INNER JOIN
+    pg_catalog.pg_class AS index_c
+    ON pg_constraint.conindid = index_c.oid
+WHERE
+    constraint_namespace.nspname = 'public'
+    AND pg_constraint.contype = 'f'
     AND pg_constraint.conislocal;
 
 -- name: GetFunctions :many
--- a/pkg/diff/schema_migration_plan_test.go
+++ b/pkg/diff/schema_migration_plan_test.go
@@ -43,150 +43,6 @@ var (
 
 	schemaMigrationPlanTestCases = []schemaMigrationPlanTestCase{
 		{
-			name: "No-op",
-			oldSchema: schema.Schema{
-				Tables: []schema.Table{
-					{
-						Name: "foobar",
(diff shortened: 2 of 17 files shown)


pkg/diff/sql_generator.go
```diff
@@ -1225,29 +1295,44 @@ func (isg *indexSQLVertexGenerator) addDepsOnTableAddAlterIfNecessary(index sche
...
+			EscapedName: schema.EscapeIdentifier(csg.tableName),
+		}, schema.EscapeIdentifier(con.Name)), con.Expression))
 	if !con.IsInheritable {
 		sb.WriteString(" NO INHERIT")
 	}
 
 	if !con.IsValid {
 		sb.WriteString(" NOT VALID")
+	} else if !csg.isNewTable {
+		hazards = append(hazards, MigrationHazard{
```
bplunkett-stripe, line 1325: We were missing this hazard previously

internal/queries/queries.sql
```diff
@@ -109,6 +111,35 @@ WHERE
...
+INNER JOIN pg_catalog.pg_namespace AS foreign_table_namespace
+    ON
+        foreign_table_c.relnamespace = foreign_table_namespace.oid
+INNER JOIN
+    pg_catalog.pg_class AS index_c
+    ON pg_constraint.conindid = index_c.oid
+WHERE
+    constraint_namespace.nspname = 'public'
+    AND pg_constraint.contype = 'f'
+    AND pg_constraint.conislocal;
```
alexaub-stripe, line 141: Does this imply that fk constraints for partitioned tables will be handled only at the parent table level?

pkg/diff/schema_migration_plan_test.go
```diff
@@ -42,150 +42,6 @@ var (
 	}
 
 	schemaMigrationPlanTestCases = []schemaMigrationPlanTestCase{
-		{
-			name: "No-op",
```
alexaub-stripe, line 46: What's the purpose of removing these test cases?

pkg/diff/sql_generator.go
```diff
@@ -402,18 +447,27 @@ func (schemaSQLGenerator) Alter(diff schemaDiff) ([]Statement, error) {
...
-	}
-	indexGraphs, err := diff.indexDiffs.resolveToSQLGraph(&indexSQLVertexGenerator)
+	})
 	if err != nil {
 		return nil, fmt.Errorf("resolving index sql graphs: %w", err)
 	}
 
+	fkConsGraphs, err := diff.foreignKeyConstraintDiffs.resolveToSQLGraph(&foreignKeyConstraintSQLVertexGenerator{
+		deletedTablesByName:           deletedTablesByName,
+		addedTablesByName:             addedTablesByName,
```
alexaub-stripe, line 463: double checking: Will this include recreated tables?

pkg/diff/sql_generator.go
```diff
@@ -1332,6 +1422,106 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+	var stmts []Statement
+	if !diff.old.IsValid && diff.new.IsValid {
+		diff.old.IsValid = diff.new.IsValid
+		// We're not keeping track of the other FK attributes, so it's easiest to ensure all other properties are equal
+		// by just modifying the old constraint diff to exclude "NOT VALID", which should make the diffs equal if no
+		// other properties have changed
+		if !strings.HasSuffix(diff.old.ConstraintDef, " NOT VALID") {
+			return nil, fmt.Errorf("expected the old constraint def to be suffixed with NOT VALID: %q", diff.old.ConstraintDef)
+		}
+		diff.old.ConstraintDef = strings.TrimSuffix(diff.old.ConstraintDef, " NOT VALID")
```
alexaub-stripe, line 1479: nit: Is this just so the cmp.Equal on 1482 works? Might make sense to just do that inline on 1482

pkg/diff/sql_generator.go
```diff
@@ -1332,6 +1422,106 @@ func (a *attachPartitionSQLVertexGenerator) GetDeleteDependencies(_ schema.Table
...
+	return deps
+}
+
+func (f *foreignKeyConstraintSQLVertexGenerator) GetDeleteDependencies(con schema.ForeignKeyConstraint) []dependency {
+	deps := []dependency{
+		mustRun(f.GetSQLVertexId(con), diffTypeDelete).before(buildTableVertexId(con.OwningTableUnescapedName), diffTypeDelete),
+		mustRun(f.GetSQLVertexId(con), diffTypeDelete).before(buildTableVertexId(con.ForeignTableUnescapedName), diffTypeDelete),
+	}
+	// This is the slightly lazy way of ensuring the foreign key constraint is added after the requisite index is
+	// built and marked as valid.
```
alexaub-stripe, line 1516:
```suggestion
	// This is the slightly lazy way of ensuring the foreign key constraint is deleted before the index it depends on is deleted
```

alexaub-stripe (approved):
Great! Really cool to see what plugging in a new feature like this is like today.

internal/queries/queries.sql
bplunkett-stripe: Yes. The foreign key will trickle down automatically for us. We don't need to do any complicated stuff like we do for indexes where we attach the foreign key

pkg/diff/schema_migration_plan_test.go
bplunkett-stripe: The schema migration plan cases test "how" the migration is done, not just the migration's corectness. The latter are tested by the acceptance tests.

Since there are many "no-op" acceptance tests that assert the generated plan is empty, we don't need this test case. I'm trying to minimize the number of these test cases, since it's tricky to maintain the before and after schemas.

pkg/diff/sql_generator.go
bplunkett-stripe: Yes! We validate this works in the test cases that change a table from partitioned to unpartitioned and vice-versa.

pkg/diff/sql_generator.go
bplunkett-stripe: This is a really good question. This is a pattern of the `Alter` functions that I'm switching too. 

The Alter functions need to keep track of what they can successfully alter and what they can't. If they can't successfully resolve a diff, then ideally they return an error to the user. 

I found the easiest way to do this is:
For each alterable difference,
- generate a statement to alter the difference
- update a copy of old to include the difference
Finally, compare old to new. If they are equal, then there were no unresolvable diffs.

With the case of invalid to valid, we need both update the old to be valid and the constraint def in order for the equality check to pass. The `strings.HasSuffix` is more of a confidence check that the constraint def follows the expected format.

alexaub-stripe (approved):

PR #54, opened 2023-08-10
cmd: make dsn a non-mandatory flag to allow use of Postgres environment variables

scottt:
### Description

`psql` and programs using `libpq` allows users to use the environment variables `PGHOST` and friends to specify the connection information. This PR implements the same behavior for `pg-schema-diff` by making the DSN flag non-mandatory.

### Motivation

- We already set the Postgres environment variables in our development and deployment environments.
- The Javascript and Golang libraries we use honor the variables.
- This change makes `pg-schema-diff` "just work" in our workflow.

### Implementation

-  pg-schema-diff's `(connFlags).parseConnConfig()` calls  `pgx.ParseConfig()`, which calls `pgconn.ParseConfigWithOptions()`

- `pgconn.ParseConfigWithOptions` already reads the Postgres environment variables.

### Testing

- I've tested the change manually on the command line and it works as expected

### DIFF
--- a/cmd/pg-schema-diff/apply_cmd.go
+++ b/cmd/pg-schema-diff/apply_cmd.go
@@ -28,7 +28,8 @@ func buildApplyCmd() *cobra.Command {
 			" (example: --allowed-hazards DELETES_DATA,INDEX_BUILD)")
 	lockTimeout := cmd.Flags().Duration("lock-timeout", 30*time.Second, "the max time to wait to acquire a lock. 0 implies no timeout")
 	cmd.RunE = func(cmd *cobra.Command, args []string) error {
-		connConfig, err := connFlags.parseConnConfig()
+		logger := log.SimpleLogger()
+		connConfig, err := connFlags.parseConnConfig(logger)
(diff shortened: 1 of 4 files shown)


bplunkett-stripe (changes requested):
This is super nice. I didn't realize pgx read from standard libpq style environment variables. Thanks for making this change!

Two small suggestions to round out the feature a bit:
- In [ParseConnConfig](https://github.com/stripe/pg-schema-diff/blob/main/cmd/pg-schema-diff/flags.go#L23), if the DSN is not set, we should warn users (by logging) that it will fallback onto libpq-style environment variables and defaults
- [This snippet of code seems unnecessary. We should probably remove it](https://github.com/stripe/pg-schema-diff/blob/main/cmd/pg-schema-diff/flags.go#L29-L33). The entire func can probably just be `return pgx.ParseConfig(*c.dsn)`

Thanks again for contributing! 🎉

bplunkett-stripe (approved):
LGTM! Solid set of changes

PR #56, opened 2023-08-10
Specify user in pgengine

bplunkett-stripe:

### Description
Before, if `PGUSER` was set on a user's machine, local tests would break because we didn't specify user. Now, the user is specified in the pgengine. This also switches the super user from the current user to `postgres`

### Motivation
Improved UX when running tests

### Testing
Unit tests

### DIFF
--- a/internal/pgengine/engine.go
+++ b/internal/pgengine/engine.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/user"
 	"path"
 	"strconv"
 	"strings"
@@ -43,7 +42,7 @@ func (c ConnectionOptions) ToDSN() string {
 }
 
 type Engine struct {
-	user *user.User
+	superuser string
 
 	// for cleanup purposes
 	process  *os.Process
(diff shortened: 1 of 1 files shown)


internal/pgengine/engine.go
```diff
@@ -102,27 +97,28 @@ func StartEngineUsingPgDir(pgDir string) (pgEngine *Engine, retErr error) {
...
-	if err := pgEngine.waitTillServingTraffic(maxConnAttemptsAtStartup, waitBetweenStartupConnAttempt); err != nil {
+	if err := pgEngine.waitTillServingTraffic(defaultMaxConnAttemptsAtStartup, defaultWaitBetweenStartupConnAttempt); err != nil {
 		return nil, fmt.Errorf("waiting till server can serve traffic: %w", err)
 	}
 
 	return pgEngine, nil
 }
 
-func initDB(currentUser *user.User, initDbPath, dbPath string) error {
+func initDB(initDbPath, dbPath string, superuser string) error {
```
alexrhee-stripe, line 119: (Nit.) If they're all string we can just do `(x, y, z string)`?

alexrhee-stripe (approved):

internal/pgengine/engine.go
bplunkett-stripe: ahhh true. This is a weird middle ground!

alexrhee-stripe (approved):

PR #58, opened 2023-08-11
fixed buggy english

ksarangmath:

### Description
This is fixing a word.

### Motivation
The incorrectly capitalized character was an eyesore in the readme, which is the first place users go to when viewing this repo.

### Testing
I have read the changed line like 20 times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ if err != nil {
 ```
 
 ## 2. Applying plan
-We leave plan application up to the user. For example, users might want to take out a session-level advisory lock if they are 
+We leave plan application up to the user. For example, a user might want to take out a session-level advisory lock if they are 
 concerned about concurrent migrations on their database. They might also want a second user to approve the plan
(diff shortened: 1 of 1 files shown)


bplunkett-stripe (approved):
Sorry for the inconvenience with this one! Thanks for Fixing this!

README.md
```diff
@@ -105,7 +105,7 @@ if err != nil {
 ```
 
 ## 2. Applying plan
-We leave plan application up to the user. For example, Users might want to take out a session-level advisory lock if they are 
+We leave plan application up to the user. For example, users might want to take out a session-level advisory lock if they are
```
Sukhmai, line 108:
```suggestion
We leave plan application up to the user. For example, a user might want to take out a session-level advisory lock if they are 
```

Sukhmai (changes requested):

bplunkett-stripe (approved):

PR #70, opened 2023-09-06
Add DOT encoding support for internal/graph

### LINKED ISSUE #38: Create way to visualize dependency graph
Create a way to visualize dependency graph. I think the best way to implement this is just to modify the Graph data structure to output a graph in some sort of standard graph format that can be ingested by a graph visualization tool, e.g. [Graphviz](https://graphviz.org/)

kevinmingtarja:
### Description
This PR adds support for encoding the dependency graph to [DOT](https://graphviz.org/doc/info/lang.html) format, which is used by [Graphviz](https://graphviz.org/), a popular graph visualization software. This will allow us to visualize our dependency graphs for debugging purposes.

### Motivation
Closes #38 

### Testing
- Added a unit test for the function
- Locally, I tested it against some schemas from the test suite and temporarily outputting the encoded graph in `resolveToSQLGraph()`, then manually visualized it using http://magjac.com/graphviz-visual-editor/ to verify the format's correctness

Example output:

I have a fresh new postgres instance and my target schema is this:
```
CREATE TABLE foobar(
    id INT,
    foo VARCHAR(255),
    bar TEXT,
    fizz INT,
    PRIMARY KEY (foo, id)
) PARTITION BY LIST (foo);
CREATE TABLE foobar_1 PARTITION OF foobar FOR VALUES IN ('foo_1');
CREATE TABLE foobar_2 PARTITION OF foobar FOR VALUES IN ('foo_2');
CREATE TABLE foobar_3 PARTITION OF foobar FOR VALUES IN ('foo_3');
```

Based on the diffs and dependency graph, The DOT encoding will output:
```
digraph G {
fontname="Helvetica,Arial,sans-serif"
node [fontname="Helvetica,Arial,sans-serif"]
edge [fontname="Helvetica,Arial,sans-serif"]
n0 [label="ADDALTER_table_foobar"]
n1 [label="ADDALTER_table_foobar_2"]
n2 [label="DELETE_table_foobar_2"]
n3 [label="ADDALTER_table_foobar_3"]
n4 [label="DELETE_table_foobar_3"]
n5 [label="DELETE_table_foobar"]
n6 [label="ADDALTER_table_foobar_1"]
n7 [label="DELETE_table_foobar_1"]
n0 -> n1
n0 -> n3
n0 -> n6
n2 -> n1
n4 -> n3
n5 -> n0
n7 -> n6
}
```
Screen Shot 2023-09-06 at 17 54 51


internal/graph/graph.go
```diff
@@ -213,3 +214,47 @@ func (g *Graph[V]) TopologicallySortWithPriority(isLowerPriority func(V, V) bool
 
 	return output, nil
 }
+
+func (g *Graph[V]) EncodeDOT(w io.Writer, sortVertices bool) (err error) {
```
bplunkett-stripe, line 218: Rather than this being a method of a `Graph`, can it instead, be a function:
`func EncodeDot(graph *Graph[V], sortVertices bool)`. It will help limit the scope of the Graph a bit. 

I'd put the function in the same file as the `DotBuilder`, since they go hand-in-hand, and maybe call the file `encode_dot.go`

Small nit: Get rid of the named parameter. This can lead to weird bugs/behavior within go and idelly should be avoided when possible

internal/graph/dotbuilder.go
```diff
@@ -0,0 +1,49 @@
+package graph
+
+import (
+	"fmt"
+	"io"
+)
+
+const attr = `fontname="Helvetica,Arial,sans-serif"
```
bplunkett-stripe, line 8: `dotLabelAttr`?

internal/graph/graph.go
```diff
@@ -213,3 +214,47 @@ func (g *Graph[V]) TopologicallySortWithPriority(isLowerPriority func(V, V) bool
...
+	builder := &dotBuilder{w}
+	err = builder.init()
```
bplunkett-stripe, line 220: Combine these lines into 
```
if err := &builder.init(); err != nil {
   return err
}
```

Alternatively, you could do the init in a constructor for the builder

internal/graph/graph.go
```diff
@@ -213,3 +214,47 @@ func (g *Graph[V]) TopologicallySortWithPriority(isLowerPriority func(V, V) bool
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = builder.finish()
```
bplunkett-stripe, line 225: Deferrals are great for closing resources that we shouldn't leave open in the event of an error, e.g., db connections, files. 

In this case, it's okay if we don't "finish()" the builder even if an error occurs. So to simplify the code a bit, just call `builder.finish()` at the end of the function rather than deferring it

internal/graph/graph.go
```diff
@@ -213,3 +214,47 @@ func (g *Graph[V]) TopologicallySortWithPriority(isLowerPriority func(V, V) bool
...
+		nodeIdsByVertex[vid] = i
+	}
+
+	for _, source := range vertexIds {
+		adjacentEdgesMap := g.edges[source]
+		for target, isAdjacent := range adjacentEdgesMap {
+			if isAdjacent {
+				err = builder.addEdge(nodeIdsByVertex[source], nodeIdsByVertex[target])
+				if err != nil {
+					return err
```
bplunkett-stripe, line 253: Maybe wrap the error with context, i.e., 'addEdge(%s, %s): %w`

internal/graph/graph_test.go
```diff
@@ -396,6 +397,44 @@ func TestTopologicallySortWithPriority(t *testing.T) {
...
+	for id, neighbors := range adjList {
+		for _, neighborId := range neighbors {
+			assert.NoError(t, g.AddEdge(id, neighborId))
+		}
+	}
+
+	// encode to DOT
+	buf := bytes.Buffer{}
+	// sort vertices to ensure deterministic ordering of nodes and edges
+	assert.NoError(t, g.EncodeDOT(&buf, true))
```
bplunkett-stripe, line 434: Move this test into `encode_dot_test.go`.

Also, do you think you could add a test for an empty graph?

bplunkett-stripe (changes requested):
The implementation here is good! I just have some small nits/tweaks.

Really cool set of changes!

internal/graph/dotbuilder.go
kevinmingtarja: Actually, we only need the node attributes since we're not writing any labels for the edges, so I'll just inline it.

internal/graph/graph_test.go
kevinmingtarja: Yeah for sure let me add a test for that.

kevinmingtarja:
Thanks for the feedback @bplunkett-stripe ! I just pushed my changes based on the review.

bplunkett-stripe (changes requested):
LGTM! Thanks for making the changes! Only small change left is getting linting to pass.

Looks like linting isn't passing -- maybe I should allow anyone who posts a PR to run the workflows? In the interim, you can use `make lint_fix` (you will need to install few things) or [you could use this docker container to identify linting errors](https://jira.corp.stripe.com/browse/RELDB-1565)

kevinmingtarja:
I just did `make lint_fix`. It should be good to go now!

bplunkett-stripe:
Nice! Let's hop back onto the ticket to figure out the structure of the CLI for this

PR #71, opened 2023-09-14
Add support for replica identities

bplunkett-stripe:

### Description
Add support for replica identities

### Motivation
(See above)

### Testing
Unit test

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -106,6 +106,15 @@ func (s Schema) Hash() (string, error) {
 	return fmt.Sprintf("%x", hashVal), nil
 }
 
+type ReplicaIdentity string
+
+const (
+	ReplicaIdentityDefault ReplicaIdentity = "d"
+	ReplicaIdentityNothing ReplicaIdentity = "n"
+	ReplicaIdentityFull    ReplicaIdentity = "f"
+	ReplicaIdentityIndex   ReplicaIdentity = "i"
+)
+
 type Extension struct {
 	SchemaQualifiedName
 	Version string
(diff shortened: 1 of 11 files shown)


internal/schema/schema.go
```diff
@@ -106,6 +106,15 @@ func (s Schema) Hash() (string, error) {
 	return fmt.Sprintf("%x", hashVal), nil
 }
 
+type ReplicaIdentity string
+
+const (
+	ReplicaIdentityDefault ReplicaIdentity = "d"
+	ReplicaIdentityNothing ReplicaIdentity = "n"
```
bplunkett-stripe, line 113: I chose to not do any sort of enum conversion in the schema extraction. This will enable us to support "new replica identities" added by Postgres assuming there is no diff in the identity. Otherwise, if postgres added some new replica identity, pg-schema-diff would always error out on it even if there was no diff

pkg/diff/sql_generator.go
```diff
@@ -689,10 +700,35 @@ func (t *tableSQLVertexGenerator) Delete(table schema.Table) ([]Statement, error
...
+		stmts = alterPartitionStmts
+	} else {
+		alterBaseTableStmts, err := t.alterBaseTable(diff)
+		if err != nil {
+			return nil, fmt.Errorf("altering base table: %w", err)
+		}
+		stmts = alterBaseTableStmts
+	}
+
+	if diff.old.ReplicaIdentity != diff.new.ReplicaIdentity {
```
bplunkett-stripe, line 720: Previously, we had the alter statement generation completely separated. Altering replica identity is the same for both partitions and base tables, so we now have a place we can put these shared alter stateemnts

pkg/diff/schema_migration_plan_test.go
```diff
@@ -265,48 +271,6 @@ var (
 				},
 			},
 		},
-		{
```
bplunkett-stripe, line 268: This was not testing the "how". Acceptance tests already test the same functionality but are easier to maintain. (We don't need this test case)

alexrhee-stripe (approved):
lgtm

PR #74, opened 2023-10-11
Add unique constraint support

### LINKED ISSUE #72: Unique constraints

bplunkett-stripe:

### Description
Add unique constraint support

### Motivation
Closes #72 
### Testing
Tested with unit tests

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -224,7 +224,7 @@ type (
 )
 
 const (
-	PkIndexConstraintType IndexConstraintType = "PRIMARY KEY"
+	PkIndexConstraintType IndexConstraintType = "p"
 )
 
 func (i Index) GetName() string {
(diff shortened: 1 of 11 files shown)


internal/schema/schema.go
```diff
@@ -545,15 +540,6 @@ func fetchIndexes(ctx context.Context, q *queries.Queries) ([]Index, error) {
 	return indexes, nil
 }
 
-func getIndexConstraintType(constraintType string) (IndexConstraintType, error) {
```
bplunkett-stripe, line 548: Don't bother converting the constraint type string into an enum. The old method would cause schema extraction to fail if a new constraint type was added -- even if it wasn't used anywhere in the diffing code.

alexaub-stripe (approved):
Looks amazing! 

thought: I wonder if we should have a hazard for operations like adding a unique constraint where adding the constraint can fail because of duplicate rows. I guess not since there's nothing functionally wrong with that outcome, and the UX wouldn't be great?

bplunkett-stripe:
> Looks amazing!
> 
> thought: I wonder if we should have a hazard for operations like adding a unique constraint where adding the constraint can fail because of duplicate rows. I guess not since there's nothing functionally wrong with that outcome, and the UX wouldn't be great?

I think that's an interesting idea! Hypothetically, if building a unique index fails, the database could be left in a somewhat dirty state. I think it's a bit of scope of this PR, but we might want to start thinking about overhauling the migration hazards a bit. Identifying:
- Does it make sense here?
- Are we missing any? Especially schema migrations that might affect correctness, but also maybe if it leaves the state of the database dirty

It might also be interesting if schema migration hazards include some list of "id" targets, such as for deltees

PR #83, opened 2024-01-09
Online not null constraints: `SET NOT NULL`

bplunkett-stripe:

### Description
Use check constraints to make `SET NOT NULL` constraints online

One exception: This does not yet work for partitioned tables, since we don't yet support local check constraints

### Motivation
https://github.com/stripe/pg-schema-diff/issues/82

### Testing
Tested via unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ $ pg-schema-diff plan --dsn "postgres://postgres:postgres@localhost:5432/postgre
   * Online index replacement: If some index is changed, the new version will be built before the old version is dropped, preventing a window where no index is backing queries
   * Online check constraint builds: Check constraints are added as `INVALID` before being validated, eliminating the need
 	for a long access-exclusive lock on the table
+  * Online `ALTER ... SET NOT NULL` using check constraints to eliminate the need for an access-exclusive lock on the table
   * Prioritized index builds: Building new indexes is always prioritized over deleting old indexes
 * A comprehensive set of features to ensure the safety of planned migrations:
   * Operators warned of dangerous operations.
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -3,6 +3,7 @@ package graph
 import (
 	"fmt"
 	"sort"
+	"strings"
 )
 
 type Vertex interface {
@@ -190,7 +191,12 @@ func (g *Graph[V]) TopologicallySortWithPriority(isLowerPriority func(V, V) bool
 			}
 		}
 		if indexOfSourceWithHighestPri == -1 {
-			return nil, fmt.Errorf("cycle detected: %+v, %+v", graph, incomingEdgeCountByVertex)
+			dotSB := strings.Builder{}
+			if err := EncodeDOT(g, &dotSB, true); err != nil {
+				dotSB.Reset()
+				dotSB.WriteString(fmt.Sprintf("failed to encode graph to DOT: %v", err))
+			}
+			return nil, fmt.Errorf("cycle detected: %+v, %+v\n%s", graph, incomingEdgeCountByVertex, dotSB.String())
 		}
 		sourceWithHighestPriority := sources[indexOfSourceWithHighestPri]
 
--- a/internal/migration_acceptance_tests/column_cases_test.go
+++ b/internal/migration_acceptance_tests/column_cases_test.go
@@ -452,10 +452,176 @@ var columnAcceptanceTestCases = []acceptanceTestCase{
 			);
 			`,
 		},
-		expectedHazardTypes: []diff.MigrationHazardType{
-			diff.MigrationHazardTypeAcquiresAccessExclusiveLock,
+		ddl: []string{
+			"ALTER TABLE \"public\".\"foobar\" ADD CONSTRAINT \"not_null_10111213-1415-4617-9819-1a1b1c1d1e1f\" CHECK(\"foobar\" IS NOT NULL) NOT VALID",
+			"ALTER TABLE \"public\".\"foobar\" VALIDATE CONSTRAINT \"not_null_10111213-1415-4617-9819-1a1b1c1d1e1f\"",
+			"ALTER TABLE \"public\".\"foobar\" ALTER COLUMN \"foobar\" SET NOT NULL",
+			"ALTER TABLE \"public\".\"foobar\" DROP CONSTRAINT \"not_null_10111213-1415-4617-9819-1a1b1c1d1e1f\"",
 		},
 	},
+	{
+		name: "Set NOT NULL (add invalid CC)",
+		oldSchemaDDL: []string{
+			`
+			CREATE TABLE foobar(
+			    id INT PRIMARY KEY,
+				foobar VARCHAR(255)
+			);
+			`,
+		},
+		newSchemaDDL: []string{
+			`
+			CREATE TABLE foobar(
+			    id INT PRIMARY KEY,
+				foobar VARCHAR(255) NOT NULL
+			);
+			ALTER TABLE foobar ADD CONSTRAINT foobar CHECK (foobar IS NOT NULL) NOT VALID;
+			`,
+		},
+		ddl: []string{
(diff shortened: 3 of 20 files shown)


README.md
```diff
@@ -45,6 +45,7 @@ $ pg-schema-diff plan --dsn "postgres://postgres:postgres@localhost:5432/postgre
   * Online index replacement: If some index is changed, the new version will be built before the old version is dropped, preventing a window where no index is backing queries
   * Online check constraint builds: Check constraints are added as `INVALID` before being validated, eliminating the need
 	for a long access-exclusive lock on the table
+  * Online `ALTER ... SET NOT NULL` using check constraints to eliminate the need for an access-exclusive lock on the table
```
bplunkett-stripe, line 48: Once this gets merged in, I'll cut a follow-up PR that highlights this behavior via an example, since I think showing rather than telling will work better.

internal/migration_acceptance_tests/column_cases_test.go
```diff
@@ -452,59 +473,253 @@ var columnAcceptanceTestCases = []acceptanceTestCase{
 			);
 			`,
 		},
-		expectedHazardTypes: []diff.MigrationHazardType{
-			diff.MigrationHazardTypeAcquiresAccessExclusiveLock,
+		ddl: []string{
+			"ALTER TABLE \"public\".\"foobar\" ADD CONSTRAINT \"not_null_10111213-1415-4617-9819-1a1b1c1d1e1f\" CHECK(\"foobar\" IS NOT NULL) NOT VALID",
+			"ALTER TABLE \"public\".\"foobar\" VALIDATE CONSTRAINT \"not_null_10111213-1415-4617-9819-1a1b1c1d1e1f\"",
+			"ALTER TABLE \"public\".\"foobar\" ALTER COLUMN \"foobar\" SET NOT NULL",
+			"ALTER TABLE \"public\".\"foobar\" DROP CONSTRAINT \"not_null_10111213-1415-4617-9819-1a1b1c1d1e1f\"",
```
alexaub-stripe, line 480: super cool

pkg/diff/sql_generator.go
```diff
@@ -1506,26 +1613,108 @@ func (csg *checkConstraintSQLGenerator) Delete(con schema.CheckConstraint) ([]St
...
+	if len(targetColumns) == 1 {
+		targetColumn := targetColumns[0]
+		if _, ok := csg.addedColumnsByName[targetColumn.Name]; !ok {
+			isOnPreExistingColumn = true
+		}
+	}
+
+	if isOnPreExistingColumn && isValidNotNullCC(con) {
+		// If the NOT NULL check constraint is on a pre-existing column, then we should ensure it is added before
+		// the column.
```
alexaub-stripe, line 1664: Is this correction right?
```suggestion
		// the column alter.
```

pkg/diff/sql_generator.go
```diff
@@ -1506,26 +1613,108 @@ func (csg *checkConstraintSQLGenerator) Delete(con schema.CheckConstraint) ([]St
...
+	deps := []dependency{
+		mustRun(csg.GetSQLVertexId(con), diffTypeDelete).before(csg.GetSQLVertexId(con), diffTypeAddAlter),
+	}
+
+	targetColumns, err := getTargetColumns(con, csg.newSchemaColumnsByName)
+	if err != nil {
+		return nil, fmt.Errorf("getting target columns: %w", err)
+	}
+
+	isOnPreExistingColumn := false
```
alexaub-stripe, line 1654: thought: Maybe make this isValidNotNullCCOnPreExistingColumn? `isOnPreExistingColumn` is confusing on its own, since its only used for single column not null constraints. Scenarios where there is a constraint on multiple existing columns but this variable is false are especially confusing.

pkg/diff/sql_generator.go
```diff
@@ -1506,26 +1613,108 @@ func (csg *checkConstraintSQLGenerator) Delete(con schema.CheckConstraint) ([]St
...
+		return nil, fmt.Errorf("getting target columns: %w", err)
+	}
+
+	for _, tc := range targetColumns {
+		// This is a weird quirk of our graph system, where if a -> b and b -> c and b does-not-exist, b will be
+		// implicitly created s.t. a -> b -> c (https://github.com/stripe/pg-schema-diff/issues/84)
+		//
+		// In this case, "a" is the deletion of the check constraint, "b" is the deletion of
+		// the column, and "c" is the alter/addition of the column. We do not want this behavior. We only want
+		// a -> b -> c iff the column is being delted.
```
alexaub-stripe, line 1688:
```suggestion
		// a -> b -> c iff the column is being deleted.
```

pkg/diff/sql_generator.go
```diff
@@ -1506,26 +1613,108 @@ func (csg *checkConstraintSQLGenerator) Delete(con schema.CheckConstraint) ([]St
...
+		if _, ok := csg.deletedColumnsByName[tc.Name]; ok {
+			deps = append(deps, mustRun(csg.GetSQLVertexId(con), diffTypeDelete).before(buildColumnVertexId(tc.Name), diffTypeDelete))
+		}
+	}
+
+	// If it's a not-null check constraint, we can drop the check constraint whenever convenient, e.g., after
+	// the column has been altered because `NOT NULL` does not depend on the type of the column.
+	// For all other check constraints, they can rely on the type of the column. Thus, we should drop these
+	// check constraint before any columns are altered because the new type might not be compatible with the old
+	// check constraint.
```
alexaub-stripe, line 1698: Is there value to allowing the NN constraints to be dropped whenever? Maybe it'd be best to add the dependency regardless just for consistency among the constraint types?

alexaub-stripe (approved):
Looks great! Some thoughts but all non-blocking.

pkg/diff/sql_generator.go
```diff
@@ -842,32 +875,90 @@ func (t *tableSQLVertexGenerator) GetAddAlterDependencies(table, _ schema.Table)
...
+	return isNotNullCCRegex.MatchString(cc.Expression)
+}
+
+func buildTempNotNullConstraint(colDiff columnDiff) (schema.CheckConstraint, error) {
+	uuid, err := uuid.NewRandom()
+	if err != nil {
+		return schema.CheckConstraint{}, fmt.Errorf("generating uuid: %w", err)
+	}
+	return schema.CheckConstraint{
+		Name:               fmt.Sprintf("not_null_%s", uuid.String()),
```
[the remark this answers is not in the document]
bplunkett-stripe: Great point!

pkg/diff/sql_generator.go
bplunkett-stripe: True. I'll try to make this more clear. I also had doubts about this var's naming.

pkg/diff/sql_generator.go
bplunkett-stripe: It's critical we drop this last, which the topographical sort does if an obj has no deps because of the scenario where we're deleting a valid not null constraint AND changing a column to NOT NULL.

I imagine the scenario above is pretty common because it would occur for a half-complete migration, where the valid constraint was added but the `SET NOT NULL` never ran

internal/graph/graph.go
```diff
@@ -190,7 +191,12 @@ func (g *Graph[V]) TopologicallySortWithPriority(isLowerPriority func(V, V) bool
 			}
 		}
 		if indexOfSourceWithHighestPri == -1 {
-			return nil, fmt.Errorf("cycle detected: %+v, %+v", graph, incomingEdgeCountByVertex)
+			dotSB := strings.Builder{}
+			if err := EncodeDOT(g, &dotSB, true); err != nil {
+				dotSB.Reset()
```
bplunkett-stripe, line 196: The dot encoding has been super useful for debugging! @kevinmingtarja

pkg/diff/sql_generator.go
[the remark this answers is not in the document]
bplunkett-stripe: I want to update the temporary indexes to use this pattern, but I'm going to save that for a subsequent PR, since I also am going to switch to the UUID's to base64 encoded to make them smaller.

alexaub-stripe (approved):

PR #91, opened 2024-01-22
Concurrent schema fetching

bplunkett-stripe:

### Description
Make schema fetching concurrent. This should accelerate it a fair bit, especially if individual queries have high latency. 

### Motivation
Accelerate schema fetching

### Testing
Tested via unit tests


internal/schema/schema.go
```diff
@@ -348,41 +350,110 @@ func (t Trigger) GetName() string {
...
 func GetPublicSchema(ctx context.Context, db queries.DBTX) (Schema, error) {
-	q := queries.New(db)
+	// To allow backwards compatibility with connections, we will not use concurrency if passed in a db that is not a
+	// *sql.DB. This is because not all implementations are thread safe, e.g., pgx.Connection.
+	//
+	// In the future, we should maybe create options where users can pass in a DB pool (WithPool(db) or WithConnection(db))
+	// and we can set concurrency to 1 if the passed in db is not a *sql.DB.
+	goRoutineRunner := util.NewSynchronousGoRoutineRunner()
+	if _, ok := db.(*sql.DB); ok {
+		goRoutineRunner = util.NewGoroutineLimiter(50)
```
bplunkett-stripe, line 360: Bound the number of go routines we start. This isn't strictly necessary, but is probably a good practice.

ericw-stripe (approved):
Nice!

internal/concurrent/future.go
```diff
@@ -0,0 +1,76 @@
+package concurrent
+
+import "context"
+
+// Future are a way to easily start a go routine and get the result of the go routine later. Avoid building any
```
alexaub-stripe, line 5:
```suggestion
// Future provides a way to easily start a go routine and get the result of the go routine later. Avoid building any
```

internal/concurrent/future.go
```diff
@@ -0,0 +1,76 @@
...
+// mapping functionality into futures, as that can make them difficult to reason about
+type (
+	result[T any] struct {
+		err error
+		res T
+	}
+
+	GoRoutineRunner interface {
+		// Go starts a go routine and returns an error if the go routine could not be started.
+		Go(context.Context, func()) error
```
alexaub-stripe, line 15: Is there ever a scenario where the goroutine cannot be started? Is it for an already canceled/completed context? (in which case SynchronousGoRoutineRunner should probably behave the same?)

internal/concurrent/future.go
```diff
@@ -0,0 +1,76 @@
...
+		return zeroVal, ctx.Err()
+	case res := <-f.resultChan:
+		// Re-send the result to the channel so that the next call to Get() will get the same computed value. This
+		// should be safe because the channel is buffered with a size of 1.
+		f.resultChan <- res
+		return res.res, res.err
+	}
+}
+
+func ResolveAll[T any](ctx context.Context, futures ...Future[T]) ([]T, error) {
```
alexaub-stripe, line 66: thought: GetAll to align with Future.Get?

internal/concurrent/future.go
alexaub-stripe: Ah I see GoRoutineLimter now... disregard ^

internal/concurrent/go_routine_limiter_test.go
```diff
@@ -0,0 +1,89 @@
+package concurrent
```
alexaub-stripe, line 1: nit: file name doesn't align with the file being tested

internal/concurrent/future.go
alexaub-stripe: Well hmm GoRoutineLimter errors on Go on canceled context, so maybe the sync & async runners should too... Let me know what you think

internal/concurrent/goroutine_limiter.go
```diff
@@ -0,0 +1,32 @@
...
+
+// NewGoroutineLimiter creates a new GoroutineRunner that limits the number of goroutines. No more than the limit number
+// of goroutines can be running at the same time.
+func NewGoroutineLimiter(limit int64) *LimitedGoRoutineRunner {
+	return &LimitedGoRoutineRunner{
+		sem: semaphore.NewWeighted(limit),
+	}
+}
+
+func (l *LimitedGoRoutineRunner) Go(ctx context.Context, fn func()) error {
```
alexaub-stripe, line 21: Should document out that this will block rather surprisingly in the middle of NewFuture rather than in Future.Get. Is that desirable -- I guess it avoids a proliferation of errors but is a pretty big break to the contract of NewFuture

alexaub-stripe (dismissed):
Overall LGTM, some thinking around the concurrency primitives

internal/concurrent/future.go
bplunkett-stripe: Good point! I'll add it!

internal/concurrent/goroutine_limiter.go
bplunkett-stripe: Good point about documentation! 

This behavior is desired albeit unclear)! The point of the GoRoutineLimiter is to limit the number of go routines we spin up. It also lets us force concurrent code to be serializable to make our schema fetching compatible with connections.

bplunkett-stripe:
>Well hmm GoRoutineLimter errors on Go on canceled context, so maybe the sync & async runners should too... Let me know what you think

Per semaphore docs:
```
// Acquire acquires the semaphore with a weight of n, blocking until resources
// are available or ctx is done. On success, returns nil. On failure, returns
// ctx.Err() and leaves the semaphore unchanged.
//
// If ctx is already done, Acquire may still succeed without blocking.
```

They only check ctx cancellation if the thread is blocked waiting on acquiring a unit.

Either way, we are left with a race condition without implementing some chan behavior, so I think it's probably _okay_ as is, albeit a little weird.

internal/concurrent/goroutine_limiter.go
alexaub-stripe: > The point of the GoRoutineLimiter is to limit the number of go routines we spin up
Is our goal to limit the number of goroutines or to limit the number of goroutines doing the expensive work? 

> It also lets us force concurrent code to be serializable to make our schema fetching compatible with connections.
Wouldn't this happen anyways if you moved the semaphore check inside the goroutine execution? It would occur during Future.Get, which I figure is the expected behavior.

bplunkett-stripe: It's too limit the number of go routines. Go supports hundreds of thousands of go routines, but it's a bad practice to leave it entirely unbounded.

> Wouldn't this happen anyways if you moved the semaphore check inside the goroutine execution? It would occur during Future.Get, which I figure is the expected behavior.
No, it wouldn't without a specialized future implementation, since futures eagerly compute. 

We would inherently be spinning up  goroutines if we didn't block on this aspect, which makes the code not thread-safe. For instance, I tried making the code serializable initially by setting go routine limit == 1, but that doesn't work because if a future calls into another a future, the shared "thread pool" will have deadlock, since 1 go routine can't complete with the other.

alexaub-stripe (approved):

Issue #94, opened 2024-01-25
Support for non-`public` schemas


markalexander:
Hi,

This is currently blocking our use-case.

Any insight into how much work this would be if we wanted to make a PR?

I.e. it just a case of adding an appropriate parameter and replacing the hard-coded instances of `public`, or is there some hidden complexity to it?  

From memory, some objects are namespaced unusually in Postgres, which might lead to conflicts?  Though I only recall having those problems with the deeper features like RLS policies.

Thanks!

bplunkett-stripe:
I totally understand! I realize this is a feature a _lot_ of folks want to, so it's next on my list!

Making this change isn't entirely trivial because, many objects like tables are namespaced by schema. So I can have `some_schema_1.foobar` and `some_schema_2.foobar`. 

pg-schema-diff builds the migration plan using a web of dependencies. Initially, we built this system to only support `public`, so a lot of internal representations of names do _not_ include the schema. As a result, it would fall apart in scenarios where there are two identically named objects in different schemas -- failing out or generating entirely wrong plans.

I plan to plug away at this problem in the next week or so! pg-schema-diff is not my primary focus, so the timeline of this feature is pretty tentative.

bplunkett-stripe:
Multi-schema support should be merged in now 🎉  The CLI still needs to be updated to actually support it!

PR #96, opened 2024-01-29
Add support for fetching non-public schemas.

bplunkett-stripe:

### Description
This only adds support for fetching non-public schemas. This does actually allow generating migration plans for non-public schemas because we will need plumb several into the sql generation code to account for schemas.

I opted to do the filtering in-memory because it is way more flexible to do the filtering in-memory rather than in the queries. We can easily support things like regex and complex nested operations in the future (unions of intersections, etc). I don't expect the filters to significantly reduce the number of schema objects being fetched, so the performance impact of doing it in-memory is minimal.

### Motivation
https://github.com/stripe/pg-schema-diff/issues/94

### Testing
Tested via unit tests

### DIFF
--- a/internal/queries/queries.sql
+++ b/internal/queries/queries.sql
@@ -2,6 +2,7 @@
 SELECT
     c.oid AS oid,
     c.relname::TEXT AS table_name,
+    table_namespace.nspname::TEXT AS table_schema_name,
     c.relreplident::TEXT AS replica_identity,
     COALESCE(parent_c.relname, '')::TEXT AS parent_table_name,
     COALESCE(parent_namespace.nspname, '')::TEXT AS parent_table_schema_name,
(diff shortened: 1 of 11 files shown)


internal/schema/schema.go
```diff
@@ -508,26 +582,30 @@ func (s *schemaFetcher) fetchTables(ctx context.Context) ([]Table, error) {
...
 		tableFutures = append(tableFutures, tableFuture)
 	}
+	tables, err := concurrent.GetAll(ctx, tableFutures...)
+	if err != nil {
+		return nil, fmt.Errorf("getting tables: %w", err)
+	}
 
-	return concurrent.GetAll(ctx, tableFutures...)
-}
+	tables = filterSliceByName(
```
bplunkett-stripe, line 597: It is way more flexible to do this filtering in-memory rather than in the query. We can easily support things like regex and complex nested operations in the future (unions of intersections, etc).

internal/queries/queries.sql
```diff
@@ -25,8 +29,8 @@ LEFT JOIN
     pg_catalog.pg_namespace AS parent_namespace
     ON parent_c.relnamespace = parent_namespace.oid
 WHERE
-    c.relnamespace
-    = (SELECT oid FROM pg_catalog.pg_namespace WHERE nspname = 'public')
+    table_namespace.nspname NOT IN ('pg_catalog', 'information_schema')
```
bplunkett-stripe, line 32: This is the same filter Postgres uses in its pg_stat_user_tables view to filter down from pg_stat_all_tables.

PR #100, opened 2024-02-01
Add exclude schemas

bplunkett-stripe:

### Description
When using temp db factories (which have metadata tables), we need to tell the schema fetcher to ignore those tables.

This also technically switches acceptance tests to fetch non-public schemas; however, none of our tests actually have non-public schemas. 

The final step for supporting non-public schemas at this point is validating everything is actually working through the acceptance tests, updating the documentation, and removing the gate.
### Motivation
https://github.com/stripe/pg-schema-diff/issues/94

### Testing
Acceptance tests

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -155,7 +155,7 @@ func (t Table) IsPartitioned() bool {
 
 // IsPartition returns whether the table is a partition.
 // It represents a mismatch in modeling because the ForValues and ParentTable are stored separately.
-// Instead, they fields be stored under the same struct as a nilable pointer and this function should be deleted
+// Instead, the fields should be stored under the same struct as a nilable pointer, and this function should be deleted.
 func (t Table) IsPartition() bool {
 	return t.ParentTable != nil
 }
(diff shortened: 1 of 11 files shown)


internal/schema/schema.go
```diff
@@ -382,34 +392,77 @@ func GetSchema(ctx context.Context, db queries.DBTX, opts ...GetSchemaOpt) (Sche
...
-func buildNameFilter(options getSchemaOptions) nameFilter {
-	if len(options.includeSchemas) == 0 {
+func buildNameFilter(options getSchemaOptions) (nameFilter, error) {
+	if intersection := intersect(options.includeSchemas, options.excludeSchemas); len(intersection) > 0 {
+		return nil, fmt.Errorf("schemas %v are both included and excluded", intersection)
+	}
+
+	includeFilter := buildIncludeSchemasFilter(options.includeSchemas)
+	excludeFilter := buildExcludeSchemasFilter(options.excludeSchemas)
+	return andNameFilter(includeFilter, excludeFilter), nil
```
bplunkett-stripe, line 414: We technically don't need to build a `excludeSchemasFilter` if there is an `includeSchemasFilter`, however I think the flow of control here is a bit simpler.

Everything will also still work if we got rid of the intersection check with this method!

pkg/tempdb/factory.go
```diff
@@ -185,12 +202,18 @@ func (o *onInstanceFactory) Create(ctx context.Context) (sql *sql.DB, dropper Dr
...
+		return nil, err
 	}
 
-	return tempDbConn, func(ctx context.Context) error {
-		_ = tempDbConn.Close()
-		return o.dropTempDatabase(ctx, tempDbName)
+	return &Database{
+		ConnPool: tempDbConn,
+		ExcludeMetadatOptions: []schema.GetSchemaOptions{
+			schema.WithExcludeSchemas(o.options.metadataSchema),
```
bplunkett-stripe, line 211: This is the core of the PR. Without this, all acceptance tests would fail because they would try to diff out the metadata table.

PR #132, opened 2024-05-28
RLS Policies

### LINKED ISSUE #111: Row security policies
Add support for row security policies

bplunkett-stripe:

### Description
- Add support for RLS Policies
- RLS policies cannot be added to individual partitions yet, but they can be put on the parent table (I want to refactor the code a bit)

Little status update: Bit of a hiatus due to other priorities. Before adding more features to schema diffing, I'd like to clean up the tests and code a bit.

cc @sweatybridge

### Motivation
Fixes #111 

### Testing
Tested via unit tests

### DIFF
--- a/internal/migration_acceptance_tests/acceptance_test.go
+++ b/internal/migration_acceptance_tests/acceptance_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/google/uuid"
@@ -42,7 +43,9 @@ type (
 	planFactory func(ctx context.Context, connPool sqldb.Queryable, tempDbFactory tempdb.Factory, newSchemaDDL []string, opts ...diff.PlanOpt) (diff.Plan, error)
 
 	acceptanceTestCase struct {
-		name         string
+		name string
+		// roles is a list of roles that should be created before the DDL is applied
+		roles        []string
 		oldSchemaDDL []string
 		newSchemaDDL []string
 
@@ -88,12 +91,16 @@ func (suite *acceptanceTestSuite) runTestCases(acceptanceTestCases []acceptanceT
 			suite.Run("vanilla", func() {
 				suite.runSubtest(tc, tc.vanillaExpectations, nil)
 			})
-			suite.Run("with data packing (and ignoring column order)", func() {
-				suite.runSubtest(tc, tc.dataPackingExpectations, []diff.PlanOpt{
-					diff.WithDataPackNewTables(),
-					diff.WithLogger(log.SimpleLogger()),
+			// Only run the data packing test if there are expectations for it. We should strip out the embedded
+			// data packing tests and make them their own tests to simplify the test suite.
+			if !reflect.DeepEqual(tc.dataPackingExpectations, expectations{}) {
+				suite.Run("with data packing (and ignoring column order)", func() {
+					suite.runSubtest(tc, tc.dataPackingExpectations, []diff.PlanOpt{
+						diff.WithDataPackNewTables(),
+						diff.WithLogger(log.SimpleLogger()),
+					})
 				})
-			})
+			}
 		})
 	}
 }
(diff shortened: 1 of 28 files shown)


internal/migration_acceptance_tests/acceptance_test.go
```diff
@@ -42,7 +43,9 @@ type (
 	planFactory func(ctx context.Context, connPool sqldb.Queryable, tempDbFactory tempdb.Factory, newSchemaDDL []string, opts ...diff.PlanOpt) (diff.Plan, error)
 
 	acceptanceTestCase struct {
-		name         string
+		name string
+		// roles is a list of roles that should be created before the DDL is applied
+		roles        []string
```
alexaub-stripe, line 48: thought: I wonder if we should make it possible to define roles in schema somehow. Since they're global I figure that would erode the "you're only applying a single schema" paradigm we have, but it feels like a shame that roles have to be defined outside of pg-schema-diff.

internal/pgengine/actions.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+)
+
+// ResetInstance attempts to reset the cluster to a clean state.
+// It deletes all cluster level objects, i.e., roles, which are not deleted
+// by dropping database(s). This can be useful for re-using a cluster for multiple tests.
+func ResetInstance(ctx context.Context, db *sql.DB) error {
```
alexaub-stripe, line 12: thought: Shouldn't this also drop databases? As is it feels like this is implemented specifically for how it's used in the AT suite.

pkg/diff/policy_sql_generator.go
```diff
@@ -0,0 +1,319 @@
...
+	var alterPolicyParts []string
+
+	if !cmp.Equal(oldCopy.AppliesTo, diff.new.AppliesTo) {
+		alterPolicyParts = append(alterPolicyParts, fmt.Sprintf("TO %s", strings.Join(diff.new.AppliesTo, ", ")))
+		oldCopy.AppliesTo = diff.new.AppliesTo
+	}
+
+	if oldCopy.UsingExpression != diff.new.UsingExpression && diff.new.UsingExpression != "" {
+		// Weirdly, you can't actually drop a "USING EXPRESSION" clause from an ALL policy even though you
+		// can have an ALL policy with only a check expression.
```
alexaub-stripe, line 232: shouldn't there be an ErrNotImplemented somewhere to cover this case?

alexaub-stripe: Ah i see that's handled by if diff.old isn't mutated to match diff.new

pkg/diff/policy_sql_generator.go
```diff
@@ -0,0 +1,319 @@
...
+	deps := []dependency{
+		mustRun(psg.GetSQLVertexId(newPolicy), diffTypeDelete).before(psg.GetSQLVertexId(newPolicy), diffTypeAddAlter),
+	}
+
+	newTargetColumns, err := getTargetColumns(newPolicy.Columns, psg.newSchemaColumnsByName)
+	if err != nil {
+		return nil, fmt.Errorf("getting target columns: %w", err)
+	}
+
+	// Run after the new columns are added/altered
```
alexaub-stripe, line 283: > Policies on partitions must be dropped before the base table is altered, otherwise  the SQL could fail because, e.g., the policy references a column that no longer exists.

Is this quirk from above covered by the column dependencies?

alexaub-stripe: Ah maybe it's handled by how the policy generator is called by the table generator?

alexaub-stripe (approved):
Some thoughts, but nothing blocking. Supporting RLS is super impressive!

also thank goodness for acceptance tests...

internal/pgengine/actions.go
bplunkett-stripe: We actually have the AT's drop the database. This is more design cluster-scope values. I could totally see databases considered cluster-scope though... Right now, we will just keep the individual AT's dropping their own databases, since if they fail to do that, it indicates a bug.

pkg/diff/policy_sql_generator.go
bplunkett-stripe: Yep. That's the "canonical" pattern I'm going for. It makes "ErrNotImplemented" free for any feature we inherently haven't explicitly coded support for.

pkg/diff/policy_sql_generator.go
bplunkett-stripe: Yes, this is a quirk covered by the column dependencies. It's also the same reason we can't easily support partition-level RLS policies yet

PR #136, opened 2024-05-30
Update Acceptance Tests - Flatten specs

### LINKED ISSUE #106: Move data packing tests into their own test cases instead of having them as a inherent subtest of acceptance tests
This will cleanup the test code a bit and will greatly accelerate our test cases, since we don't really need to test data packing with each test case.

bplunkett-stripe:

### Description
- Instead of each test case defining two specs for vanilla and data packing, run data packing as their own independent test cases. This simplifies test specs and running tess
- Flatten `expectations`. The distinction was arbitrary because of the need to potentially define two test specs

The fact that this is net negative code despite copying two test cases -- we really only had TWO data packing test cases for all of this complexity in the test spec -- implies this PR moves the code in the right direction.
### Motivation
Fixes #106 
### Testing
Covered by tests

### DIFF
--- a/internal/migration_acceptance_tests/acceptance_test.go
+++ b/internal/migration_acceptance_tests/acceptance_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"reflect"
 	"testing"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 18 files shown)


alexaub-stripe (approved):
Nice!

PR #139, opened 2024-06-22
fix: #67 converting all tabs to spaces in migration acceptance tests

### LINKED ISSUE #67: Fix Test Case Indent Formatting
A lot of the test cases have inconsistent indenting formatting in the SQL. Some uses tabs, some use spaces. As a result, when this is rendered in github, some lines appear more indented than others. 

To fix this, we will need to replace all test cases that have tabs indenting with a visual equivalent in spaces. 

I'm guessing our go formatting didn't spot this because the SQL is contained within a string, which is presumably ignored by the formatter.

ammiranda:

### Description

Standardizing the whitespace in the migration_acceptance_tests to be spaces so viewing isn't uneven.

### Motivation

Resolves #67 

### Testing

I ran the test docker image built on my local machine and observed no test failures which is expected given the only changes are whitespace. The one caveat is it was only run against `postgres` 14 given that was the specified arg in the test dockerfile.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ If you want to make changes yourself, follow these steps:
 2. Make your changes
 3. Test your changes
 ```bash
- docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test --build-arg POSTGRES_PACKAGE=postgresql{14, 15} .
+ docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test .
  ```
 3. Submit a [pull request](https://help.github.com/articles/creating-a-pull-request-from-a-fork/)
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,10 @@ If you want to make changes yourself, follow these steps:
 2. Make your changes
 3. Test your changes
 ```bash
- docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test .
+# builds image running tests with the postgresql14 image (can also omit because it is the default arg)
+ docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test --build-arg POSTGRES_PACKAGE=postgresql14 .
+# builds image running tests with the postgresql15 image.
+ docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test --build-arg POSTGRES_PACKAGE=postgresql15 .
  ```
 3. Submit a [pull request](https://help.github.com/articles/creating-a-pull-request-from-a-fork/)
 
--- a/internal/migration_acceptance_tests/acceptance_test.go
+++ b/internal/migration_acceptance_tests/acceptance_test.go
@@ -1,264 +1,264 @@
 package migration_acceptance_tests
 
 import (
-    "context"
-    "database/sql"
-    "fmt"
-    "testing"
+	"context"
+	"database/sql"
+	"fmt"
+	"testing"
 
-    "github.com/google/uuid"
-    _ "github.com/jackc/pgx/v4/stdlib"
-    "github.com/kr/pretty"
-    "github.com/stretchr/testify/suite"
-    "github.com/stripe/pg-schema-diff/internal/pgdump"
-    "github.com/stripe/pg-schema-diff/internal/pgengine"
-    "github.com/stripe/pg-schema-diff/pkg/diff"
-    "github.com/stripe/pg-schema-diff/pkg/log"
-    "github.com/stripe/pg-schema-diff/pkg/sqldb"
+	"github.com/google/uuid"
+	_ "github.com/jackc/pgx/v4/stdlib"
+	"github.com/kr/pretty"
+	"github.com/stretchr/testify/suite"
+	"github.com/stripe/pg-schema-diff/internal/pgdump"
+	"github.com/stripe/pg-schema-diff/internal/pgengine"
+	"github.com/stripe/pg-schema-diff/pkg/diff"
+	"github.com/stripe/pg-schema-diff/pkg/log"
+	"github.com/stripe/pg-schema-diff/pkg/sqldb"
 
-    "github.com/stripe/pg-schema-diff/pkg/tempdb"
+	"github.com/stripe/pg-schema-diff/pkg/tempdb"
 )
 
 var (
-    errValidatingPlan = fmt.Errorf("validating migration plan")
+	errValidatingPlan = fmt.Errorf("validating migration plan")
 )
 
 type (
-    planFactory func(ctx context.Context, connPool sqldb.Queryable, tempDbFactory tempdb.Factory, newSchemaDDL []string, opts ...diff.PlanOpt) (diff.Plan, error)
+	planFactory func(ctx context.Context, connPool sqldb.Queryable, tempDbFactory tempdb.Factory, newSchemaDDL []string, opts ...diff.PlanOpt) (diff.Plan, error)
 
-    acceptanceTestCase struct {
-        name string
+	acceptanceTestCase struct {
+		name string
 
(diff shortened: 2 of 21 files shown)


CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ If you want to make changes yourself, follow these steps:
 2. Make your changes
 3. Test your changes
 ```bash
- docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test --build-arg POSTGRES_PACKAGE=postgresql{14, 15} .
+ docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test .
```
ammiranda, line 20: The command that was written here would fail because the `Dockerfile.test` file already specifies an `ARG` so I wanted to update this so others are not lead astray.

bplunkett-stripe: I'm not sure this is true. A The [build arg ](https://docs.docker.com/reference/dockerfile/#arg)specified is the default build arg if one is not specified. 

```
docker build -t pg-schema-diff-test-runner14 -f ./build/Dockerfile.test --build-arg POSTGRES_PACKAGE=postgresql15
```
works for me

bplunkett-stripe (changes requested):
The change to the contributing.md I believe is not correct. 

Also, only the SQL strings should be changed. The actual code should not have substitutions made. The goal here is to consistent spacing on the SQL strings, since the linter does not enforce anything in them.

CONTRIBUTING.md
ammiranda: That command also works for me if I include the current directory arg i.e. `.` when in the root of the repo. The original command still does not work, should I update the doc to the command in your comment?

CONTRIBUTING.md
```diff
@@ -17,7 +17,10 @@ If you want to make changes yourself, follow these steps:
 2. Make your changes
 3. Test your changes
 ```bash
- docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test --build-arg POSTGRES_PACKAGE=postgresql{14, 15} .
+# builds image running tests with the postgresql14 image (can also omit because it is the default arg)
```
ammiranda, line 20:
```suggestion
# builds image running tests with the postgresql14 image (can also omit `build-arg` because postgresql14 is the default)
```

CONTRIBUTING.md
ammiranda: I updated the contributing to be a bit clearer, both commands are confirmed to build.

bplunkett-stripe (commented):
@ammiranda, the code is still formatted with tabs instead of spaces. Please do not change the actual code. You can see the workflows lint failing. You can use `make lint` to validate this (once you install the deps). 

The goal of this is to get consistent formatting in the SQL strings, not change any actual go code, which is already being linted and automatically formatted.

Feel free to re-add me to the PR once you fix that.

ammiranda:
@bplunkett-stripe ok I think the latest push is getting closer to what is desired. One thing that was interesting is if I converted the tabs to spaces for the opening ` in the first few files the lint would give `goimport-ed` failures, if they are left alone the lint passes. Any guidance on what to do for this is appreciated because the strings do kind of look wonky, I could adjust it so the tab replacement doesn't happen for the lines that solely have the ` characters but don't know if that is acceptable. Also I achieved this through a python script that I can add into the source so it can be run in the future easily so just let me know if I should push it up.

bplunkett-stripe:
>ok I think the latest push is getting closer to what is desired. One thing that was interesting is if I converted the tabs to spaces for the opening ` in the first few files the lint would give goimport-ed failures, if they are left alone the lint passes. Any guidance on what to do for this is appreciated because it feels a bit wonky, I could adjust it so the tab replacement doesn't happen for the lines that solely have the ` characters but don't know if that is acceptable. Also I achieved this through a python script that I can add into the source so it can be run in the future easily (possibly in its own docker container) so just let me know if I should push it up.

Hmmm I see....Instead of converting tabs to spaces, what if we converted spaces to tabs? That would also align with go formatting recommendation and give us consistent white space throughout. I'd be interested in seeing the script!

ammiranda:
> > ok I think the latest push is getting closer to what is desired. One thing that was interesting is if I converted the tabs to spaces for the opening `in the first few files the lint would give goimport-ed failures, if they are left alone the lint passes. Any guidance on what to do for this is appreciated because it feels a bit wonky, I could adjust it so the tab replacement doesn't happen for the lines that solely have the` characters but don't know if that is acceptable. Also I achieved this through a python script that I can add into the source so it can be run in the future easily (possibly in its own docker container) so just let me know if I should push it up.
> 
> Hmmm I see....Instead of converting tabs to spaces, what if we converted spaces to tabs? That would also align with go formatting recommendation and give us consistent white space throughout. I'd be interested in seeing the script!

@bplunkett-stripe I just pushed up the change changing 4 spaces to tabs. You can view the python script as a gist here: https://gist.github.com/ammiranda/44c720878f09aad21e8e17e305988537. Let me know if you want me to write a Dockerfile to enable it to be easily ran as part of this PR!

bplunkett-stripe:
> @bplunkett-stripe I just pushed up the change changing 4 spaces to tabs. You can view the python script as a gist here: https://gist.github.com/ammiranda/44c720878f09aad21e8e17e305988537. Let me know if you want me to write a Dockerfile to enable it to be easily run as part of this PR!

Nice changes, look solid!  Thinking about it a bit more...I think we want to convert the tabs to spaces, i.e., reverse your script! Sorry for the confusion for that, but since you wrote a script, hopefully a pretty easy thing to switch up! Should be good to merge after that.


Thanks for the script! might look into vendoring your script in later. I'm a little hesitant since the matching condition is pretty loose.

ammiranda:
@bplunkett-stripe I've switched it back to spaces, please review again when able. If you can let me know how you'd want me to tighten the script I can work on implementing it, it seems to work for the existing formatting FWIW but I understand the desire to not be sloppy.

bplunkett-stripe (approved):
Nice seems good to merge in. In terms of the script, I think it should first be converted to  golang to standardize the # of languages we have the repo.

In terms of aligning with the other linters, it probably makes sense to have a `--fix` flag, i.e., a validation mode vs a write mode. 

Alternatively, you can have it always write to the files, and we can add it to the code_gen step.

ammiranda:
Got it, if you create an issue with your specifications I can work on that separately in a new PR. Or if you want me to create it I can do that.

bplunkett-stripe:
Doing as a separate PR is probably best. I think what we're aiming for here is to just have a golang "script" that:
- Re-writes the acceptance tests to have tab-less multi-line strings
- Preferably has a`--fix` option, such that there is basically a validation/dry-run mode and a write mode
- Is called from the makefile `lint` and `lint_fix` targets (like go_lint, sql_lint, etc)

bplunkett-stripe:
@ammiranda Are you planning on working on a go-version of your script as described above?

ammiranda:
@bplunkett-stripe sure! I can create the issue and you can update it and add details as needed if that works for you.

bplunkett-stripe:
> @bplunkett-stripe sure! I can create the issue and you can update it and add details as needed if that works for you.
Sounds good with me, thanks!

bplunkett-stripe:
```python
import os
import in_place

dir_path = "./internal/migration_acceptance_tests"

filenames = os.listdir(dir_path)

for filename in filenames:
    path = os.path.join(dir_path, filename)
    if os.path.isdir(path):
        continue
    with in_place.InPlace(path) as fp:
        in_sql_string = False
        for line in fp:
            if in_sql_string and '`' not in line:
                new_line = line.replace("\t", "    ")
                fp.write(new_line)
            else:
                fp.write(line)
            if '`' in line:
                in_sql_string = not in_sql_string

```

Updated place-holder script to account for dirs

PR #146, opened 2024-07-18
Add test cases for functional indexes

alexaub-stripe:

### Description
Adds test cases to cover [functional indexes](https://www.postgresql.org/docs/7.3/indexes-functional.html) to ensure that we test them in perpetuity.

A callout: under the hood the apply generates with a type cast for the input to the lower function (which only accepts text not varchar) which is different from the user provided definition without the cast. I think this doesn't trip up the tests and is desirable behavior because in reality pg-schema-diff is diffing schema that's been applied and read from a test database, not the user provided schema, so that type cast is always present.

Let me know if those empty line cleanup diffs should be removed. That's something my editor did.

### Motivation
We want to use functional indexes. pg-schema-diff already does, but it's preferable to ensure it's covered by tests.

### Testing
CI

### DIFF
--- a/internal/migration_acceptance_tests/index_cases_test.go
+++ b/internal/migration_acceptance_tests/index_cases_test.go
@@ -106,6 +106,31 @@ var indexAcceptanceTestCases = []acceptanceTestCase{
 		},
 	},
 	{
+		name: "Add a functional index",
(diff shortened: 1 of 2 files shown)


bplunkett-stripe (approved):
Nice

PR #147, opened 2024-07-19
Fix column renames for indexes

bplunkett-stripe:

### Description
Index columns effectively have two occurrences in pg_attribute, one is a "virtual column" created for the index and the other is the actual referenced column. For our purposes, we care about the latter, i.e., ensuring an index is deleted before its owning column is dropped, otherwise the index will be implicitly deleted (and not concurrently).

When a column is renamed, the virtual column name stays the same, but the underlying column name obviously changes. pg-schema-diff contained a bug where it referenced the _virtual_ column name. Effectively, a column would be renamed but the column names in the fetched index would not change, leading to incorrectly generated dependencies.

[I found this thread in the Postgres maling list to be very useful.
](https://www.postgresql.org/message-id/20210224.165511.465857914180630454.horikyota.ntt%40gmail.com)

What I did:
- Updated pg-schema-diff to reference the actual column name
- (added benefit) Index columns are now fetched in the same query as the indexes, reducing round trips to the database

### Motivation
Bug fix

### Testing
See added test case

### DIFF
--- a/internal/queries/queries.sql
+++ b/internal/queries/queries.sql
@@ -95,6 +95,13 @@ SELECT
     i.indisunique AS index_is_unique,
     COALESCE(parent_c.relname, '')::TEXT AS parent_index_name,
     COALESCE(parent_namespace.nspname, '')::TEXT AS parent_index_schema_name,
+    (
+        SELECT ARRAY_AGG(att.attname ORDER BY indkey_ord.ord)
+        FROM UNNEST(i.indkey) WITH ORDINALITY AS indkey_ord (attnum, ord)
+        INNER JOIN
+            pg_catalog.pg_attribute AS att
+            ON att.attrelid = table_c.oid AND att.attnum = indkey_ord.attnum
+    )::TEXT [] AS column_names,
     COALESCE(con.conislocal, false) AS constraint_is_local
 FROM pg_catalog.pg_class AS c
 INNER JOIN pg_catalog.pg_index AS i ON (i.indexrelid = c.oid)
@@ -118,14 +125,6 @@ WHERE
     AND table_namespace.nspname !~ '^pg_toast'
     AND table_namespace.nspname !~ '^pg_temp'
     AND (c.relkind = 'i' OR c.relkind = 'I');
-
--- name: GetColumnsForIndex :many
-SELECT a.attname::TEXT AS column_name
-FROM pg_catalog.pg_attribute AS a
-WHERE
-    a.attrelid = $1
-    AND a.attnum > 0
-ORDER BY a.attnum;
 
 -- name: GetCheckConstraints :many
 SELECT
--- a/internal/schema/schema_test.go
+++ b/internal/schema/schema_test.go
@@ -893,6 +893,38 @@ var (
 			},
 		},
 		{
+			name: "Column rename",
+			ddl: []string{`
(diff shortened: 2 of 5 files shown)


pkg/diff/sql_generator.go
```diff
@@ -1718,6 +1718,7 @@ func (isg *indexSQLVertexGenerator) addDepsOnTableAddAlterIfNecessary(index sche
 		}
 	}
 
+	// Otherwise, we can drop the index whenever we want.
```
bplunkett-stripe, line 1721: Initially was debugging this problem and thought adding a comment helped document it a little better

internal/schema/schema_test.go
```diff
@@ -892,6 +892,38 @@ var (
...
+							{Name: "renamed_value", Type: "text", IsNullable: true, Size: -1, Collation: defaultCollation},
+						},
+						CheckConstraints: nil,
+						ReplicaIdentity:  ReplicaIdentityDefault,
+					},
+				},
+				Indexes: []Index{
+					{
+						OwningTable: SchemaQualifiedName{SchemaName: "public", EscapedName: "\"foo\""},
+						Name:        "foo_value_idx", Columns: []string{"renamed_value"},
```
bplunkett-stripe, line 921: Previously, it would have fetched the column as "value" instead of "renamed_value"

internal/queries/queries.sql
```diff
@@ -95,6 +95,13 @@ SELECT
     i.indisunique AS index_is_unique,
     COALESCE(parent_c.relname, '')::TEXT AS parent_index_name,
     COALESCE(parent_namespace.nspname, '')::TEXT AS parent_index_schema_name,
+    (
+        SELECT ARRAY_AGG(att.attname ORDER BY indkey_ord.ord)
+        FROM UNNEST(i.indkey) WITH ORDINALITY AS indkey_ord (attnum, ord)
```
alexaub-stripe, line 100: The ordinality here is just to make it explicit that the unnested array should be aggregated back in the same order after converting to column names right? And this isn't technically necessary since that should be the order anyways?

internal/queries/queries.sql
```diff
@@ -95,6 +95,13 @@ SELECT
     i.indisunique AS index_is_unique,
     COALESCE(parent_c.relname, '')::TEXT AS parent_index_name,
     COALESCE(parent_namespace.nspname, '')::TEXT AS parent_index_schema_name,
+    (
+        SELECT ARRAY_AGG(att.attname ORDER BY indkey_ord.ord)
```
alexaub-stripe, line 99: I almost wonder if it's worth a comment on this block here that this is just `i.indkey.map(attnum -> column_name(attnum))` but in sql.

alexaub-stripe (approved):
Amazing catch

internal/queries/queries.sql
bplunkett-stripe: Yep exactly. It's to enforce the order. I don't think it's technically necessary, but I definitely feel more comfortable with explicit `ORDER BY`

internal/queries/queries.sql
bplunkett-stripe: Might leave out the comment, since it's probably best to read from the docs on it. Pretty nifty SQL function!

PR #151, opened 2024-07-24
Fix Properly terminate sessions for temporary database to allow deletion

KEHyeon:
### Description

Because of the database environment, the actual database is designated as a temporary database, so when a session is connected and the database is deleted, an error occurs, so code is added to terminate the session connected to the temporary database and delete the database.

### Motivation
[//]: # Eliminate problems that occur when deleting temporary databases

### Testing
[//]: # Tested locally

### DIFF
--- a/internal/queries/queries.sql
+++ b/internal/queries/queries.sql
@@ -95,6 +95,13 @@ SELECT
     i.indisunique AS index_is_unique,
     COALESCE(parent_c.relname, '')::TEXT AS parent_index_name,
     COALESCE(parent_namespace.nspname, '')::TEXT AS parent_index_schema_name,
+    (
+        SELECT ARRAY_AGG(att.attname ORDER BY indkey_ord.ord)
+        FROM UNNEST(i.indkey) WITH ORDINALITY AS indkey_ord (attnum, ord)
+        INNER JOIN
+            pg_catalog.pg_attribute AS att
+            ON att.attrelid = table_c.oid AND att.attnum = indkey_ord.attnum
+    )::TEXT [] AS column_names,
     COALESCE(con.conislocal, false) AS constraint_is_local
 FROM pg_catalog.pg_class AS c
 INNER JOIN pg_catalog.pg_index AS i ON (i.indexrelid = c.oid)
@@ -118,14 +125,6 @@ WHERE
     AND table_namespace.nspname !~ '^pg_toast'
     AND table_namespace.nspname !~ '^pg_temp'
     AND (c.relkind = 'i' OR c.relkind = 'I');
-
--- name: GetColumnsForIndex :many
-SELECT a.attname::TEXT AS column_name
-FROM pg_catalog.pg_attribute AS a
-WHERE
-    a.attrelid = $1
-    AND a.attnum > 0
-ORDER BY a.attnum;
 
 -- name: GetCheckConstraints :many
 SELECT
--- a/internal/queries/queries.sql.go
+++ b/internal/queries/queries.sql.go
@@ -87,38 +87,6 @@ func (q *Queries) GetCheckConstraints(ctx context.Context) ([]GetCheckConstraint
 	return items, nil
 }
 
-const getColumnsForIndex = `-- name: GetColumnsForIndex :many
-SELECT a.attname::TEXT AS column_name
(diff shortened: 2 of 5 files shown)


bplunkett-stripe:
I'm not sure we want to force terminate connections because it would indicate someone is still using the database, which might imply we have incorrectly identified a temporary database (a bug), or some other aspect of the code is bugged.

I want to better understand your use case to identify how to best structure this type of change! Are you using the tempdb package outside of the CLI?

Thanks for helping out with pg-schema-diff! 🎉

KEHyeon:
We are currently developing a program to check whether the schemas of the DB operated by our service and the sql files we manage match. So, if multiple people access the db while running the program, a session connection is created in tempdb, an error remains when dropping, and tempdb remains in the running db. To solve this, I think you can unconditionally drop tempdb. This is because tempdb is not a DB that will be used anyway, but is a DB created only for program operation. If there is anything you don't understand, please let me know. Thank you!

KEHyeon:
Additionally, checking session information can lead to tools such as dbeaver. I think this problem arises when you develop a program that actually uses DB like this.

bplunkett-stripe:
> We are currently developing a program to check whether the schemas of the DB operated by our service and the sql files we manage match. So, if multiple people access the db while running the program, a session connection is created in tempdb, an error remains when dropping, and tempdb remains in the running db. To solve this, I think you can unconditionally drop tempdb. This is because tempdb is not a DB that will be used anyway, but is a DB created only for program operation. If there is anything you don't understand, please let me know. Thank you!

Could the solution maybe be terminating the connections on your end of the temp db? You could build a connection pool to your database and issue a similar query.

For what you are developing, you could use pg-schema-diff to build a migration plan against your database and the SQL files. If the plan is empty, then your schema matches. You could pass a connection to pg-schema-diff that acquires a Postgres Advisory Lock before it generates the migration plan, preventing other clients from doing anything.

KEHyeon:
The explanation was insufficient. We are using a DB to compare with tempdb. This is to make settings such as timesaledb and role the same. I think that's why such an issue occurred. Why can't terminating the tempdb session be the solution?

KEHyeon:
The explanation was insufficient. We are using fromdb as tempdb. This is to make settings such as timesaledb and role the same. I think that's why such an issue occurred. Why can't terminating the tempdb session be the solution?

bplunkett-stripe:
Terminating connections on the database implies connections are being leaked by your client. As in, after you set your settings on your temp db, can you not close the connection?

This is how I imagine your flow to be:
1. Create tempdb
2. Open connection on temp db
3. Set your various settings
4. Close connection on temp db (this should mean it's now possible to kill the tempdb without forcing connections closed)
5. Pass temp db connection pool into pg-schema-diff
6. Close temp db

This shouldn't require force killing any connections unless there is a bug in your code or my own. Correct me if I'm wrong here! I'm

KEHyeon:
sorry for late reply. Currently, we are using the same tempdb DB as the actual deployed DB, so I think problems arise because multiple people can access it using tools such as dbeaver. The reason why they are used together is to meet various conditions such as timescaledb and role.

Lastly, I'm not good at English, so I may not understand well. I'm sorry.

KEHyeon:
제가 잘못이해한거 같아요. 영어를 잘못해요. 죄송합니다. 나중에 다시 올게여.

bplunkett-stripe:
> I think I misunderstood. sorry. Im not good at English

No worries, that's totally fine!

Let me know if you think what I explained above answers your question. Feel free to ask any follow-up questions you might have!

KEHyeon:
thank you What I am talking about now is a different topic than above.  After applying a query to tempdb, there is work I want to do using 
image
like this

bplunkett-stripe:
You could hypothetically just build your own `tempdb.Factory` implementation (it's an [interface](https://github.com/stripe/pg-schema-diff/blob/main/pkg/tempdb/factory.go#L50-L56)) via composition and wire that behavior in yourself.

What are you trying to do with the tempdb factory? I'm trying to determine if other folks would benefit from a similar behavior, or if it's pretty use-case specific.

KEHyeon:
I'm trying to check the data entered into tempdb through an insert query. So, I created an option to access tempdb in the middle.

PR #152, opened 2024-08-02
Feat: add DoSomthingOnTempdb

KEHyeon:
### Description
[//]: #After queries are executed in tempdb, an option has been added to allow work using tempdb.

### Motivation
[//]: # The applied DML cannot be checked using pg-schema-diff. However, I needed a function to check the DML applied to tempdb.  So, I wanted to add the ability to intercept tempdb and check the data in it after applying a query to tempdb. This is my motivation for implementing that option.

### Testing
[//]: # tested locally

thank you!


KEHyeon:
실수를 해서 다시만들게요..

PR #153, opened 2024-08-02
Feat: add DoSomthingOnTempdb option

KEHyeon:
Description
[//]: #After queries are executed in tempdb, an option has been added to allow work using tempdb.

Motivation
[//]: # The applied DML cannot be checked using pg-schema-diff. However, I needed a function to check the DML applied to tempdb. So, I wanted to add the ability to intercept tempdb and check the data in it after applying a query to tempdb. This is my motivation for implementing that option.

Testing
[//]: # tested locally

thank you!

### DIFF
--- a/cmd/pg-schema-diff/plan_cmd.go
+++ b/cmd/pg-schema-diff/plan_cmd.go
@@ -76,7 +76,7 @@ type (
 	}
 
 	schemaSourceFlags struct {
-		schemaDir         string
+		schemaDirs        []string
 		targetDatabaseDSN string
 	}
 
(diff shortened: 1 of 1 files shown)


bplunkett-stripe:
I wonder if it makes more sense for your project to define a custom schema source rather than vendoring this into pg-schema-diff? This seems pretty use-case specific, so I'm hesitant to merge this in. It's a good idea, but merging in niche-case puts creates more of a maintenance burden on the project.

KEHyeon:
Then, how about modifying ddlSchemaSource like this?

KEHyeon:
The PR was closed due to a mistake.  How about adding the doSomthingOnTempDb interface to ddlSchemaSource?

KEHyeon:
tempdb에 쿼리문을 실행후 그 값을 이용하기 위해 어떤 식으로 tempdbfactory를 구현해야 할까요?

KEHyeon:
How should tempdbfactory be implemented to use the value after executing a query in tempdb?

bplunkett-stripe:
So sorry for the late response @KEHyeon ! So this is how I think we can achieve what you want:
```
1. Create your own temp db using the temp db factory
2. Run your DDL and DML against the temp db
3. Pass a `diff.DbSchemaSource(tempDb)` to the `diff.Generate` function (remember to also pass in the `tempDb.ExcludeMetadataOptions`
```

This effectively will enable you to create an arbitrary schema source without the library needing to export schema yet. Let me know if that helps!

PR #164, opened 2024-08-30
Add support for column identities

### LINKED ISSUE #163: Add support for identity columns
```
-- Table: dbo.Url

-- DROP TABLE IF EXISTS dbo.Url;

CREATE TABLE dbo."Url"
(
    "URL_KEY" BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
    "DOMAIN" VARCHAR(100) NOT NULL,
    "PATH" VARCHAR(1000) NOT NULL,
    "HASH" VARCHAR(50) UNIQUE
);

CREATE INDEX ix_url_ix_hash
    ON dbo."Url" USING btree ("HASH");
```

creates 

Screenshot 2024-08-29 at 1 20 34 AM

where as it should create 

```
-- Table: dbo.Url

-- DROP TABLE IF EXISTS dbo."Url";

CREATE TABLE IF NOT EXISTS dbo."Url"
(
    "URL_KEY" bigint NOT NULL GENERATED BY DEFAULT AS IDENTITY ( INCREMENT 1 START 1 MINVALUE 1 MAXVALUE 9223372036854775807 CACHE 1 ),
    "DOMAIN" character varying(100) COLLATE pg_catalog."default" NOT NULL,
    "PATH" character varying(1000) COLLATE pg_catalog."default" NOT NULL,
    "HASH" character varying(50) COLLATE pg_catalog."default",
    CONSTRAINT "Url_pkey" PRIMARY KEY ("URL_KEY"),
    CONSTRAINT "Url_HASH_key" UNIQUE ("HASH")
)

TABLESPACE pg_default;

ALTER TABLE IF EXISTS dbo."Url"
    OWNER to postgres;
-- Index: ix_url_ix_hash

-- DROP INDEX IF EXISTS dbo.ix_url_ix_hash;

CREATE INDEX IF NOT EXISTS ix_url_ix_hash
    ON dbo."Url" USING btree
    ("HASH" COLLATE pg_catalog."default" ASC NULLS LAST)
    TABLESPACE pg_default;
```

bplunkett-stripe:

### Description
Add support for column identities. This is different from generated columns.

### Motivation
Fixes #163

### Testing
Passes CI


internal/schema/schema.go
```diff
@@ -822,6 +841,19 @@ func (s *schemaFetcher) buildTable(
 			}
 		}
 
+		var identity *ColumnIdentity
+		if len(column.IdentityType) > 0 {
+			identity = &ColumnIdentity{
+				Type:       ColumnIdentityType(column.IdentityType),
```
alexaub-stripe, line 847: Should there be an error check here in case an unexpected type is returned?

alexaub-stripe (approved):
LGTM! One nit

internal/schema/schema.go
bplunkett-stripe: No! This is very intentional. This enables us to diff schemas without erroring out if the column identity doesn 't change. You could imagine Postgres adds a new column identity. If we errored here, our tooling would totally break even if the column doesn't change. Instead, we should aim to error out within the SQL generator, i.e., only if what we don't know about **changes**

Issue #129, opened 2024-05-06
Better dependency tracking for non-sql functions


ewan-escience:
When trying out this product with the `plan` command, the program errors, because it seems to do some live validation of the plan. If I understand correctly what is happening, this generates errors because:

1. some of the function are written in `plpgsql`, for which the program doesn't check its dependencies
2. those functions seem to be created in alphabetical order, so some functions are created before their dependencies

One of the following would solve this:
1. add flag is added that sets `check_function_bodies = off` before doing the validation
2. add support for dependency checking of `plpgsql` functions
3. don't validate the plan when the plan was created from two live databases (i.e. with both `--dsn` and `--schema-source-dsn`)

bplunkett-stripe:
Yes, that is an unfortunate limitation of postgres. It doesn't actually build any hard dependencies with pgplsql functions, so pg-schema-diff is kind of operating in the dark. We could manually try to parse dependencies from the function but that is a lot of complexity: not only would the function itself need to be parsed, but the context of the function would need to be known (is it used in a trigger, check constraint, etc). 

Plan validaiton is a confidence check. If plan validation fails, it normally implies the plan will fail against your actual database. We can definitely add an option to disable it, but I'd be curious to know if the plan validation is actually failing here, or you're okay with an invalid plan.

Thanks for the great breakdown here! I might rename the ticket to "Improve track of plpgsql functions"

ewan-escience:
Thanks for the quick reply. I'm trying to move away from [migra](https://github.com/djrobstep/migra), as it isn't developed anymore and doesn't support diffing [domains](https://www.postgresql.org/docs/current/sql-createdomain.html). Migra would always add `set check_function_bodies = off;` to its diffs, and I never had any problems with it.

I'm quite sure the plan would work, but I cannot test it, as no plan is generated at all.

How does the validation work? Is it executed in the database to upgrade? Is it isolated? Is it reverted afterwards? I would like to disable it, as we always manually check the diff anyways.

bplunkett-stripe:
I added a flag to disable plan validation in the cli via `--disable-plan-validation`.

The way plan validation works:
1. Spins up a temporary database on your postgres instance
2. Places the current schema in the temporary database
3. Runs the migration
4. Re-generates a diff and asserts that the diff is empty


We unfortunately do not support diffing of domains yet. If you add a ticket, we can try to get to it, since I imagine implementing that won't be too tricky.

peterldowns:
[...]

PR #168, opened 2024-09-02
feat: add --pre-plan-file flag

aleclarson:

### Description

Add `--pre-plan-file` argument. It should point to a SQL file. Its statements are executed before the migration plan statements.

### Motivation

This is particularly useful for settings (e.g. `SET` statements) that need to be applied before the migration plan. I'm especially interested in running `SET check_function_bodies = off;` as suggested in https://github.com/stripe/pg-schema-diff/issues/129#issuecomment-2096033386.

### Testing

None yet.

### Notes

You have my permission to take this PR and run with it! :)

### DIFF
--- a/cmd/pg-schema-diff/apply_cmd.go
+++ b/cmd/pg-schema-diff/apply_cmd.go
@@ -122,6 +122,14 @@ func runPlan(ctx context.Context, connConfig *pgx.ConnConfig, plan diff.Plan) er
 	}
 	defer conn.Close()
 
+	if plan.PrePlanDDL != "" {
+		fmt.Println(header("Executing pre-plan DDL"))
+		fmt.Printf("%s\n\n", plan.PrePlanDDL)
+		if _, err := conn.ExecContext(ctx, plan.PrePlanDDL); err != nil {
+			return fmt.Errorf("executing pre-plan DDL: %w", err)
+		}
+	}
+
 	// Due to the way *sql.Db works, when a statement_timeout is set for the session, it will NOT reset
 	// by default when it's returned to the pool.
 	//
--- a/cmd/pg-schema-diff/plan_cmd.go
+++ b/cmd/pg-schema-diff/plan_cmd.go
@@ -77,6 +77,7 @@ type (
 
 	schemaSourceFlags struct {
 		schemaDirs        []string
+		prePlanFile       string
 		targetDatabaseDSN string
 	}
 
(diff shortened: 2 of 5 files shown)


aleclarson:
I've tested this locally and it's working as expected :)

bplunkett-stripe:
In order to get this to pass plan validation, you will need to wire it through the plan generator as well (into the library and not just the CLI). I can also try to pick this up later if you can wait a bit.

aleclarson:
I have it running the pre-plan file on the temp DB now as well. For some reason, the function is still being checked (and therefore failing with `type "foo" does not exist (SQLSTATE 42704)` as it did before.

aleclarson:
I've looked into this a bit more. Here's what I found.

- While `check_function_bodies = off` helps, it's not the whole fix.
- The problem is that I'm using `RETURNS SETOF foo` in my function signature, which requires the table to exist at the time of `CREATE FUNCTION` execution.

I'm devising a workaround that may be of interest to you. Here's how it works.

- Before generating the migration plan, replace `SETOF foo` with `TABLE (_id int)`. The columns don't need to match the `foo` table's columns, because I'm setting `check_function_bodies = off`.
- After running the migration plan, my script runs, dropping each function that needed the placeholder return type. Then it runs the `CREATE FUNCTION` statements for those functions.

I will let you know how it goes! 🤞

aleclarson:
I've had to implement a `--pre-apply-file` flag as well (not present in this PR). Here's why:

- Using a pre-plan file is great for things like `SET check_function_bodies = OFF;` where no database objects are created.
- Creating a database object with a pre-plan file is unadviseable, because it messes with the temporary database (used for plan validation) in a way that breaks it.
- Why am I trying to create a database object with a pre-plan file? Good question. My tool is handling edge cases that `pg-schema-diff` doesn't support yet, like `CREATE TYPE` statements.
- So I devised a new “pre-apply file” that gets executed everywhere the pre-plan file does **except** when preparing the temporary database's initial schema.
- Along with using the pre-apply file, my tool executes `DROP TYPE … CASCADE` statements when a change is detected. This is done before ever running `pg-schema-diff`. This is one part of ensuring that the temporary database won't contain any database objects that I'm creating in the pre-apply file.

aleclarson:
I won't have time to finish this one

aleclarson:
@bplunkett-stripe Can't say for sure. I ended up taking a totally different approach that involves parsing the SQL myself, doing a topological sort, creating any objects that don't yet exist, and only then do I call pg-schema-diff (with plan validation disabled). This is working well for me, which means the pre-plan file is no longer needed --in my case--.

bplunkett-stripe:
>@bplunkett-stripe Can't say for sure. I ended up taking a totally different approach that involves parsing the SQL myself, doing a topological sort, creating any objects that don't yet exist, and only then do I call pg-schema-diff (with plan validation disabled). This is working well for me, which means the pre-plan file is no longer needed --in my case--.

Ahh okay. I'm sorry you had to create that kind of work around!

Issue #131, opened 2024-05-27
Make statement generation and ordering independent


bplunkett-stripe:
To do this, every generated statement should be a SQL node. Instead of flattening sql graphs in nested sql generators, we leave them as sql graphs and union them into the larger graph. 

Every sql statement should have a globally unique id that can be referenced.

Issue #101, opened 2024-02-02
Cleanup mudballs


bplunkett-stripe:
- Split SQL generators into their own files
- Split schema diff building into its own file
- Split the schema struct definitions into a separate file from where the schemas are fetched

PR #176, opened 2024-09-27
Refactor SQL generation - graph-first approach

bplunkett-stripe:

### Description
Refactor SQL generation such that SQL generators return nested graphs instead of flattened lists of statements. This allows nested hierarchies to be diffed while having root level SQL generators build dependencies on those nested sql generators.

Follow-up PRs:
1. Leverage this primitive in the table sql generator such that column statements can be referenced from root-level sql generators
2. Make a proof-of-concept by improving the policy sql generators to depend on individual column statements
3. Re-organize code per #101 
4. Implement view support

### Motivation
https://github.com/stripe/pg-schema-diff/issues/131

### Testing
Acceptance tests pass


pkg/diff/sql_graph.go
```diff
@@ -65,3 +60,216 @@ func (s *sqlGraph) toOrderedStatements() ([]Statement, error) {
...
+}
+
+func addVertexIfNotExists(graph *sqlGraph, vertex sqlVertex) {
+	if !graph.HasVertexWithId(vertex.GetId()) {
+		graph.AddVertex(vertex)
+	}
+}
+
+type sqlVertexGenerator[S schema.Object, Diff diff[S]] interface {
+	Add(S) (partialSQLGraph, error)
```
bplunkett-stripe, line 142: Now, SQL generators can return entire graphs for any one operation. I.e. when a table is altered, it can return n addresable SQL vertex nodes, one for each column, that SQL vertex generators will be able to globally reference

pkg/diff/sql_vertex_generator.go
```diff
@@ -0,0 +1,229 @@
...
+			priority:   sqlPriorityUnset,
+			statements: nil,
+		}
+		targetVertex := sqlVertex{
+			id:         dep.target,
+			priority:   sqlPriorityUnset,
+			statements: nil,
+		}
+
+		// To maintain the correctness of the graph, we will add a dummy vertex for the missing dependencies
```
bplunkett-stripe, line 67: We probably want to remove this behavior in the future, but that's out-of-scope of this PR

pkg/diff/sql_graph.go
```diff
@@ -4,64 +4,124 @@ import (
...
-	return fmt.Sprintf("%s_%s", s.DiffType, s.ObjId)
+	return s.id.String()
 }
 
-func buildTableVertexId(name schema.SchemaQualifiedName) string {
-	return fmt.Sprintf("table_%s", name)
+func (s sqlVertex) GetPriority() int {
+	// Prioritize adds/alters over deletes. Weight by number of statements. A 0 statement delete should be
+	// prioritized over a 1 statement delete
+	return len(s.statements) * int(s.priority)
```
alexaub-stripe, line 66: Statement count probably shouldn't be able to override a configured priority which would be surprising. I would only expect statement count to break ties. Some options:
* Expose a comparison function for sqlVertex instead of GetPriority, so you can order things first by priority then by statement count
* Otherwise find some scheme to encode priority and statement count losslessly in an integer (ie priority * 1000 + len(statements), though we'd need a limit on statement count)

pkg/diff/sql_graph.go
```diff
@@ -4,64 +4,124 @@ import (
...
 
-func buildIndexVertexId(name schema.SchemaQualifiedName) string {
-	return fmt.Sprintf("index_%s", name)
+// dependency indicates an edge between the SQL to resolve a diff for a source schema object and the SQL to resolve
+// the diff of a target schema object
+//
+// Most SchemaObjects will have two nodes in the SQL graph: a node for delete SQL and a node for add/alter SQL.
+// These nodes will almost always be present in the sqlGraph even if the schema object is not being deleted (or added/altered).
+// If a node is present for a schema object where the "diffType" is NOT occurring, it will just be a no-op (no SQl statements)
+type dependency struct {
```
alexaub-stripe, line 75: nit: a doc on whether source depends on target or vice versa would be helpful

pkg/diff/sql_graph.go
```diff
@@ -4,64 +4,124 @@ import (
...
-		DiffType:   old.DiffType,
-		Statements: append(old.Statements, new.Statements...),
+func (d dependencyBuilder) after(id sqlVertexId) dependency {
+	return dependency{
+		source: id,
+		target: d.base,
+	}
+}
+
+// sqlGraph represents two dependency webs of SQL statements
```
alexaub-stripe, line 104: nit: "two" graphs? Where does that come from?

pkg/diff/sql_vertex_generator.go
```diff
@@ -0,0 +1,229 @@
...
+		vertices:     vertices,
+		dependencies: dependencies,
+	}
+}
+
+func graphFromPartials(parts partialSQLGraph) (*sqlGraph, error) {
+	graph := newSqlGraph()
+	for _, vertex := range parts.vertices {
+		// It's possible the node already exists. merge it if it does
+		if graph.HasVertexWithId(vertex.GetId()) {
```
alexaub-stripe, line 49: We also need to remove the vertex that's already in there along with adding the merged one right?

pkg/diff/sql_vertex_generator.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (s *partialSQLGraph) statements() []Statement {
+	var statements []Statement
+	for _, vertex := range s.vertices {
+		statements = append(statements, vertex.statements...)
+	}
+	return statements
+}
+
+func concatPartialGraphs(parts ...partialSQLGraph) partialSQLGraph {
```
alexaub-stripe, line 32: thought: It might be nice to get rid of concatPartialGraphs and just have graphFromPartials take a list of partials. That way we could potentially add validation on partial construction (ie no dupe deps or vertices). It also might be nice to have each separated partial right up until the point where we merge them.

pkg/diff/sql_vertex_generator.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		targetVertex := sqlVertex{
+			id:         dep.target,
+			priority:   sqlPriorityUnset,
+			statements: nil,
+		}
+
+		// To maintain the correctness of the graph, we will add a dummy vertex for the missing dependencies
+		addVertexIfNotExists(graph, sourceVertex)
+		addVertexIfNotExists(graph, targetVertex)
```
alexaub-stripe, line 69: nit: I think if you just make addVertexIfNotExists take an id instead of a whole vertex you don't need the empty vertex boilerplate above

pkg/diff/sql_vertex_generator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	return graph, nil
+}
+
+func mergeVertices(old, new sqlVertex) sqlVertex {
+	priority := old.priority
+	if old.priority == sqlPriorityUnset {
+		priority = new.priority
+	}
```
alexaub-stripe, line 83: thought: Take the highest priority instead of taking an arbitrary one based on order of vertexes?

alexaub-stripe (changes requested):
Looking good on the approach! Some comments but nothing huge

pkg/diff/sql_vertex_generator.go
bplunkett-stripe: I don't think we can get rid of `concatPartialGraphs` because there may be some "nested SQL generators", where we would have concat multiple partials into a bigger partial.
- TableSQLGenerator -> (ColumnSQLGenerator | CheckConstraintSQLGenerator | etc). Each returns a partial that would unioned, still be a partial, because they might have dependenciecs on higher level SQL generators

pkg/diff/sql_vertex_generator.go
bplunkett-stripe: Nope, it will just override the existing vertex. This is the [exact same behavior as before.](https://github.com/stripe/pg-schema-diff/blob/main/internal/graph/graph.go#L28-L29)

pkg/diff/sql_vertex_generator.go
bplunkett-stripe: Good call

pkg/diff/sql_vertex_generator.go
bplunkett-stripe: Yes that makes sense! This just maintains the old behavior! "highest probability excluding the unset probability", i.e., `max(lowestProbablity, unset) -> 0`, which is not what we want.

pkg/diff/sql_graph.go
```diff
@@ -4,64 +4,124 @@ import (
...
 )
 
+// sqlVertexId is an interface for a vertex id in the SQL graph
+type sqlVertexId interface {
+	fmt.Stringer
+}
+
+// sqlPriority is an enum for the priority of a statement in the SQL graph, i.e., whether it should be run sooner
+// or later in the topological sort of the graph
+type sqlPriority int
```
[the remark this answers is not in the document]
bplunkett-stripe: Prioritize index adds ahead of any sort of index delete (not index replacement). We could alternatively build hard dependencies on all index deletes such that they depend on index deletes, but that's definitely out of scope.

Imagine the following:
User adds index `Foobar (created_at, id)` and deletes index `Fizzbuzz(id)`. These two indexes are totally independent, but it's important that `Foobar` is created before `Fizzbuzz`. It is much easier to use priority than build hard dependencies, albeit the latter is probably more robust.

BLUF:
- Used to prioritize indexes adds over indexes deletes
- Fixing this with dependencies is definitely out-of-scope for this PR. This system existed before the refactor

pkg/diff/sql_graph.go
bplunkett-stripe: >Statement count probably shouldn't be able to override a configured priority which would be surprising. I would only expect statement count to break ties.

The way it's coded, that's exactly what it does. It breaks ties.

pkg/diff/sql_graph.go
bplunkett-stripe: oops, that long gone

pkg/diff/sql_graph.go
bplunkett-stripe: oohh good call

pkg/diff/sql_vertex_generator.go
bplunkett-stripe: One thing I like about concating partial graphs as we go is it follows a sort-of builder pattern. I think it's a bit more confusing if you have to carry local variables tens of lines down to be included in one function.

bplunkett-stripe:
Made the suggested changes with exception to sql priority and concatting partial graphs. See my comments for explanations of the current behavior!

pkg/diff/sql_vertex_generator.go
alexaub-stripe: I figure that sqlgenerators could return lists of partials, but yeah that's fair enough. I'm not too opinionated about this.

alexaub-stripe (approved):
Responses sgtm, and change look great!

PR #189, opened 2025-01-28
Change the Generate function to take a SchemaSource instead of a sqldb.Queryable

the-glu:

### Description

This PR change the `Generate` function to take a `SchemaSource` instead of a `sqldb.Queryable`.

### Motivation
We're using this library to generate migrations and this option allows us to generate forward **and** backward migrations, based on the wanted schema and database.

The change allows for better flexibility in  `Generate` and let's us just swap arguments to the function, to generate both migrations at once.

### Testing
Tests with the `pg-schema-diff-test-runner` are passing locally :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ if err != nil {
 }
 defer tempDbFactory.Close()
 // Generate the migration plan
-plan, err := diff.Generate(ctx, connPool, diff.DDLSchemaSource(ddl),
+plan, err := diff.Generate(ctx, diff.DBSchemaSource(connPool), diff.DDLSchemaSource(ddl),
 	diff.WithTempDbFactory(tempDbFactory),
 	diff.WithDataPackNewTables(),
 )
--- a/cmd/pg-schema-diff/plan_cmd.go
+++ b/cmd/pg-schema-diff/plan_cmd.go
@@ -410,7 +410,9 @@ func generatePlan(ctx context.Context, logger log.Logger, connConfig *pgx.ConnCo
 		defer schemaSourceCloser.Close()
 	}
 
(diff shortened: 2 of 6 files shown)


bplunkett-stripe (changes requested):
Thanks for the contribution! I fundamentally agree with what you're trying to do...but maybe think `WithReverse` is a bit confusing and indicates a bigger flaw with the pg-schema-diff API.



Depending on how much work you want to do here:
# With code change
Replace `sqldb.Queryable` with `SchemaSource` and change `fromDB` to `fromSchema`. This will break backwards compatibility, but I think that's okay(?)

# No Code Change
- Use the `tempdbfactory` to create a database with your target schema DDL and pass that as `sqldb.Queryable`
- Use the database schema source as the target schema

Let me know if you're interested in the more in-depth code change! I could also try to throw this together!

the-glu:
Thanks for having a look!

> Let me know if you're interested in the more in-depth code change! I could also try to throw this together!

Ok, I can also go with the code change if you're OK with breaking backwards compatibility :)

Any specific hints or direction for the change? I see there is already a `DBSchemaSource`, so it should be doable without big modifications.

the-glu:
The PR has been updated with the new code change :)

bplunkett-stripe (changes requested):
Looks great! Could you update the code example in the README.md?

the-glu:
> Looks great! Could you update the code example in the README.md?

Yes, it should be fixed now :)

bplunkett-stripe (approved):
Thanks for updating the `README.md`! Changes look good to me. Appreciate the contribution!

the-glu:
Thanks you for the quick review and the approval :)

PR #190, opened 2025-02-05
Exclude tables (and related indexes) owned by extensions

the-glu:

### Description
This PR is similar to #127, but for tables and indexes generated by extensions.

### Motivation
I ran into issues with postgis who is creating such tables, this will just ingore them as they are handled by the normal create/delete extension statements.

### Testing
Tested locally with postgis :)


bplunkett-stripe (commented):
Awesome! See comment about testing

the-glu:
Hum, unfortunately postgis is dependent on the latest postgres version, when I add it to the Docker file, postgres16 is installed / used, even in `POSTGRES_PACKAGE=postgresql14` mode.

I checked for build-in extensions that would add some tables in the public schema, but didn't find any obvious ones.

Would it be fine to skip tests on that one? #127 had the same issue ^^'

internal/queries/queries.sql.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by sqlc. DO NOT EDIT.
 // versions:
-//   sqlc v1.18.0
+//   sqlc v1.28.0
```
bplunkett-stripe, line 3:
```suggestion
//   sqlc v1.18.0
```

bplunkett-stripe:
Looks like you need to re-generate the queries! It seems like your local sqlc is a different version, so you only need to commit probably a change in version number. You can alternatively install sqlc 1.18 or mount the docker image to your file system!

bplunkett-stripe (approved):
Sweet! Going to go ahead and merge!

Issue #191, opened 2025-02-06
Dependency missed with a function used as default value


the-glu:
Hello,

I have an issue with the following schema:

```
CREATE FUNCTION public.demo() RETURNS text
    LANGUAGE plpgsql
    AS $$
declare
begin
   return 'hi';
end;

$$;

CREATE TABLE public.demo2 (
    demo3 text DEFAULT public.demo() NOT NULL
);

```

The dependency between demo2 and demo is not detected, validation is not passing.

```
validating migration plan: running migration plan: executing migration statement: {DROP FUNCTION "public"."demo"() 3s 3s [HAS_UNTRACKABLE_DEPENDENCIES: Dependencies, i.e. other functions used in the function body, of non-sql functions cannot be tracked. As a result, we cannot guarantee that function dependencies are ordered properly relative to this statement. For drops, this means you need to ensure that all functions this function depends on are dropped after this statement.]}: ERROR: cannot drop function demo() because other objects depend on it (SQLSTATE 2BP01)
```

With generated statements lookings like this:

```
        {
            DDL:         "DROP FUNCTION \"public\".\"demo\"()",
            Timeout:     3000000000,
            LockTimeout: 3000000000,
            Hazards:     {
                {Type:"HAS_UNTRACKABLE_DEPENDENCIES", Message:"Dependencies, i.e. other functions used in the function body, of non-sql functions cannot be tracked. As a result, we cannot guarantee that function dependencies are ordered properly relative to this statement. For drops, this means you need to ensure that all functions this function depends on are dropped after this statement."},
            },
        },
        {
            DDL:         "DROP TABLE \"public\".\"demo2\"",
            Timeout:     1200000000000,
            LockTimeout: 3000000000,
            Hazards:     {
                {Type:"DELETES_DATA", Message:"Deletes all rows in the table (and the table itself)"},
            },
        },
```

I do understand that there are complexities tracking dependencies of functions, but could a 'obvious' dependency like that one, when the function is explicitly used somewhere else be detected?

I tried to dig a into the code but I was not sure where / how to find relevents part, if you have some pointers I would be happy to make a PR to improve that :)

PR #192, opened 2025-02-06
Handle columns who use functions as defaults

the-glu:

### Description
Detect and add as dependency functions that can be used as default.

I'm really not sure I'm not missing some locations where I should process the new `DependsOnFunctions` field, but it's working for my case ;)

### Motivation
See #191 for the base case issue, where default value use functions ;)

### Testing
Tested locally, tests yet to be added :)


pkg/diff/sql_generator.go
```diff
@@ -1407,13 +1421,26 @@ func buildColumnVertexId(columnName string, diffType diffType) sqlVertexId {
...
 
-func (csg *columnSQLVertexGenerator) GetDeleteDependencies(_ schema.Column) ([]dependency, error) {
-	return nil, nil
+func (csg *columnSQLVertexGenerator) GetDeleteDependencies(col schema.Column) ([]dependency, error) {
+
+	var deps []dependency
+	for _, depFunction := range col.DependsOnFunctions {
+		deps = append(deps, mustRun(csg.GetSQLVertexId(col, diffTypeDelete)).before(buildFunctionVertexId(depFunction, diffTypeDelete)))
+	}
+	return deps, nil
```
bplunkett-stripe, line 1443: I believe this does nothing right now, because the functions exist in the root graph. Ideally, we merge the graphs, but that's too much effort right now.

pkg/diff/sql_generator.go
```diff
@@ -1074,6 +1074,13 @@ func (t *tableSQLVertexGenerator) GetAddAlterDependencies(table, _ schema.Table)
 			mustRun(t.GetSQLVertexId(table, diffTypeAddAlter)).after(buildTableVertexId(*table.ParentTable, diffTypeAddAlter)),
 		)
 	}
+
+	for _, col := range table.Columns {
```
bplunkett-stripe, line 1078: This is the currently awkward part about column sql generation: add/delete of columns is pushed under table add/alter.

As a result, what I believe you need to do:
- If the column is new/altered, then you need to build a dependency between the table and the function such that function **add/alter** comes **BEFORE** the table **add/alter**
- If the column is being deleted, then you need to build a dependency between the table and the function such that the function **deletes** comes **BEFORE** the table **add/alter**
^
The same is true if the table is being totally created or totally deleted.

bplunkett-stripe (changes requested):
Thanks for taking a stab at this!

Some changes to make:
- Update the dependency generation
- Update the `schema_test` to ensure the functions are correctly fetched
- [Create acceptance tests for this](https://github.com/stripe/pg-schema-diff/tree/b0ba6873555a80904caf38587597d7d0d242023d/internal/migration_acceptance_tests) and validate the existing acceptance tests all pass. These will be your best friend. The code is fragile; the tests are robust.

pkg/diff/sql_generator.go
the-glu: Yes, but that what I'm doing no ?

Here we're in table add/alter, and we add a dep for each column's function that have a dependency, and just bellow the delete (in opposite mode)

`diffTypeDelete` may be wrong there however ^^'

pkg/diff/sql_generator.go
the-glu: So they should be removed ?

the-glu (commented):
Sorry I took me so long to be able to fix that one, I had to focus on something else ^^'

I added some acceptance tests (in the function_cases_test) that do pass.

I'm not sure I understood what I would update in `schema_test`. Should I add the base test case to be sure it's handled?

I left my question about dependency generation on individual comments as well.

the-glu:
Hi @bplunkett-stripe 

Just a gentle reminder about this PR, when you have a moment, could you please take a quick look?

Thanks a lot!

tonycosentini:
@bplunkett-stripe is there anything preventing this from ending up in the tool? Not being able to modify columns that use functions as defaults is a fairly big limitations (at least in my use cases).

bplunkett-stripe:
> @bplunkett-stripe is there anything preventing this from ending up in the tool? Not being able to modify columns that use functions as defaults is a fairly big limitations (at least in my use cases).

I'll prioritize getting this merged in.

PR #194, opened 2025-03-02
feat: add support for custom template databases

munjalpatel:
### Description
This PR introduces a new flag, `--template-db`, allowing users to specify the template database used when creating a temporary database. Previously, `template0` was hard-coded, limiting the ability to validate migrations in scenarios requiring a different template DB.

### Motivation
We frequently use various Postgres extensions that rely on different template databases for initialization, and we want to leverage migration validation against a temporary database. The prior hard-coded `template0` prevented using migration validation with those extensions. By making the template configurable, we can accommodate more specialized Postgres environments and maintain robust validation flows.

### Testing
- Added unit tests to validate the new `--template-db` flag.
- Verified downstream acceptance tests pass without regressions.
- Confirmed correct inheritance of objects from the user-specified template DB in the newly created temporary database.

### DIFF
--- a/cmd/pg-schema-diff/plan_cmd.go
+++ b/cmd/pg-schema-diff/plan_cmd.go
@@ -94,6 +94,8 @@ type (
 		lockTimeoutModifiers      []string
 		insertStatements          []string
 		outputFormat              outputFormat
+
+		templateDb string
 	}
 
 	timeoutModifier struct {
@@ -117,6 +119,8 @@ type (
 		statementTimeoutModifiers []timeoutModifier
 		lockTimeoutModifiers      []timeoutModifier
 		insertStatements          []insertStatement
+
+		templateDb string
 	}
 )
 
@@ -172,6 +176,13 @@ func createPlanFlags(cmd *cobra.Command) *planFlags {
 	)
 
 	cmd.Flags().Var(&flags.outputFormat, "output-format", "Change the output format for what is printed. Defaults to pretty-printed human-readable output. (options: pretty, json)")
+
+	cmd.Flags().StringVar(
+		&flags.templateDb,
+		"template-db",
+		"template0",
+		"Template database to use when creating temporary databases (default 'template0')",
+	)
 
 	return flags
 }
@@ -250,6 +261,7 @@ func parsePlanConfig(p planFlags) (planConfig, error) {
 		statementTimeoutModifiers: statementTimeoutModifiers,
 		lockTimeoutModifiers:      lockTimeoutModifiers,
 		insertStatements:          insertStatements,
+		templateDb:                p.templateDb,
 	}, nil
 }
 
@@ -384,7 +396,7 @@ func generatePlan(ctx context.Context, logger log.Logger, connConfig *pgx.ConnCo
 		copiedConfig := connConfig.Copy()
 		copiedConfig.Database = dbName
 		return openDbWithPgxConfig(copiedConfig)
-	}, tempdb.WithRootDatabase(connConfig.Database))
+	}, tempdb.WithRootDatabase(connConfig.Database), tempdb.WithTemplateDatabase(planConfig.templateDb))
 	if err != nil {
 		return diff.Plan{}, err
 	}
--- a/cmd/pg-schema-diff/plan_cmd_test.go
+++ b/cmd/pg-schema-diff/plan_cmd_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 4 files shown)


cmd/pg-schema-diff/plan_cmd_test.go
```diff
@@ -32,11 +33,11 @@ func TestParseTimeoutModifierStr(t *testing.T) {
 		},
 		{
 			opt:                 "timeout=15m",
-			expectedErrContains: "could not find key",
+			expectedErrContains: "could not find key", // "pattern" missing
```
bplunkett-stripe, line 36: nit: can we remove these. I understand why they were included, but I prefer not to have these comments, since they might quickly become outdated

cmd/pg-schema-diff/plan_cmd_test.go
```diff
@@ -118,3 +119,34 @@ func TestParseInsertStatementStr(t *testing.T) {
 		})
 	}
 }
+
+func TestPlanFlagsTemplateDBDefault(t *testing.T) {
```
bplunkett-stripe, line 123: It's probably okay to not include these, since we don't have analog tests. The functionality it is testing is pretty straightforward too, i.e., flag parsing

pkg/tempdb/factory.go
```diff
@@ -126,11 +135,12 @@ type (
...
-		metadataSchema: DefaultOnInstanceMetadataSchema,
-		metadataTable:  DefaultOnInstanceMetadataTable,
-		rootDatabase:   "postgres",
-		logger:         log.SimpleLogger(),
+		dbPrefix:         DefaultOnInstanceDbPrefix,
+		metadataSchema:   DefaultOnInstanceMetadataSchema,
+		metadataTable:    DefaultOnInstanceMetadataTable,
+		rootDatabase:     "postgres",
+		logger:           log.SimpleLogger(),
+		templateDatabase: DefaultTemplateDatabase,
```
bplunkett-stripe, line 143: Instead of making this a parameter of the factory, let's make this a parameter of the `Create` function, since one factory doesn't need to be restricted to one type of template

cmd/pg-schema-diff/plan_cmd.go
```diff
@@ -173,6 +177,13 @@ func createPlanFlags(cmd *cobra.Command) *planFlags {
 
 	cmd.Flags().Var(&flags.outputFormat, "output-format", "Change the output format for what is printed. Defaults to pretty-printed human-readable output. (options: pretty, json)")
 
+	cmd.Flags().StringVar(
+		&flags.templateDb,
+		"template-db",
+		"template0",
+		"Template database to use when creating temporary databases",
```
bplunkett-stripe, line 184: Maybe add to this doc string: "This is used to derive implicit schema objects for the target schema"

cmd/pg-schema-diff/plan_cmd.go
```diff
@@ -173,6 +177,13 @@ func createPlanFlags(cmd *cobra.Command) *planFlags {
 
 	cmd.Flags().Var(&flags.outputFormat, "output-format", "Change the output format for what is printed. Defaults to pretty-printed human-readable output. (options: pretty, json)")
 
+	cmd.Flags().StringVar(
+		&flags.templateDb,
+		"template-db",
```
bplunkett-stripe, line 182: nit: name `schema-template-db`

bplunkett-stripe (changes requested):
Looks really solid to me! Some small suggestions. Thanks for making these changes!

munjalpatel:
@bplunkett-stripe Thanks for the suggestions on this PR! I’m happy to incorporate improvements.

However, in my testing, I noticed an issue: when we create a new database from a template that already has extensions preloaded, and our SQL migrations don’t include `CREATE EXTENSION …` statements, `pg-schema-diff` tries to remove those extensions.

Because of that, I’m not sure how much value this PR provides unless there’s a more straightforward way to ignore everything inherited from the template database (tables, functions, extensions, etc.). Do you have any thoughts on how we might address this?

bplunkett-stripe:
> @bplunkett-stripe Thanks for the suggestions on this PR! I’m happy to incorporate improvements.
> 
> However, in my testing, I noticed an issue: when we create a new database from a template that already has extensions preloaded, and our SQL migrations don’t include `CREATE EXTENSION …` statements, `pg-schema-diff` tries to remove those extensions.
> 
> Because of that, I’m not sure how much value this PR provides unless there’s a more straightforward way to ignore everything inherited from the template database (tables, functions, extensions, etc.). Do you have any thoughts on how we might address this?

Hmmm I'm not sure why that would be occurring. Could you:
- pg_dump your template database (schema only)
- pg_dump your current database created from the template (schema only)
- Send your target ddl
- Send the command you're running and it's output

Issue #198, opened 2025-03-30
[FEATURE] Add support for diffing two schema dirs from the CLI


bplunkett-stripe:
**Describe the feature**
Add support for diffing two schemas dirs from the CLI, i.e., diffing two separate folders.

**Motivation**
https://github.com/stripe/pg-schema-diff/issues/195

PR #199, opened 2025-03-30
[FEATURE] Add support for `--from-dir` and providing a dedicated temporary database server

bplunkett-stripe:

### Description
Changes:
- Allow users to generate plans FROM a schema directory
- Allow both the from and to databases to be schema directories IF the user provides a temporary database
- Switch commands to print through the `cobra.Command` proxy
- Spin out tests for the plan and apply commands. Testing these manually was getting too complex

### Motivation
https://github.com/stripe/pg-schema-diff/issues/198

### Testing
Added tests for these core commands because it was getting too complicated to run manually. This does not all cover all cases, but it covers core usages.

### DIFF
--- a/cmd/pg-schema-diff/apply_cmd.go
+++ b/cmd/pg-schema-diff/apply_cmd.go
@@ -49,7 +49,7 @@ func buildApplyCmd() *cobra.Command {
 
 		cmd.SilenceUsage = true
 
-		plan, err := generatePlan(context.Background(), generatePlanParameters{
+		plan, err := generatePlan(cmd.Context(), generatePlanParameters{
 			fromSchema:       fromSchema,
 			toSchema:         toSchema,
 			tempDbConnConfig: connConfig,
@@ -59,12 +59,12 @@ func buildApplyCmd() *cobra.Command {
 		if err != nil {
 			return err
 		} else if len(plan.Statements) == 0 {
-			fmt.Println("Schema matches expected. No plan generated")
+			cmd.Println("Schema matches expected. No plan generated")
 			return nil
 		}
 
-		fmt.Println(header("Review plan"))
-		fmt.Print(planToPrettyS(plan), "\n\n")
+		cmd.Println(header("Review plan"))
+		cmd.Print(planToPrettyS(plan), "\n\n")
 
 		if err := failIfHazardsNotAllowed(plan, *allowedHazardsTypesStrs); err != nil {
 			return err
@@ -81,10 +81,10 @@ func buildApplyCmd() *cobra.Command {
 			}
 		}
 
-		if err := runPlan(context.Background(), connConfig, plan); err != nil {
+		if err := runPlan(cmd.Context(), cmd, connConfig, plan); err != nil {
 			return err
 		}
-		fmt.Println("Schema applied successfully")
+		cmd.Println("Schema applied successfully")
 		return nil
 	}
 
@@ -123,7 +123,7 @@ func failIfHazardsNotAllowed(plan diff.Plan, allowedHazardsTypesStrs []string) e
 	return nil
 }
 
-func runPlan(ctx context.Context, connConfig *pgx.ConnConfig, plan diff.Plan) error {
+func runPlan(ctx context.Context, cmd *cobra.Command, connConfig *pgx.ConnConfig, plan diff.Plan) error {
 	connPool, err := openDbWithPgxConfig(connConfig)
 	if err != nil {
 		return err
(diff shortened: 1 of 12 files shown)


cmd/pg-schema-diff/plan_cmd.go
```diff
@@ -38,62 +38,91 @@ func buildPlanCmd() *cobra.Command {
...
+			tempDbConnConfig: tempDbConnConfig,
+			planOptions:      planOpts,
+			logger:           logger,
+		})
 		if err != nil {
 			return err
 		}
 
-		fmt.Println(planFlags.outputFormat.convertToOutputString(plan))
+		cmd.Println(outputFmt.convertToOutputString(plan))
```
bplunkett-stripe, line 102: Canonically, all logs should be done through the command. We were just doing this wrong previously. This is how cobra does it's own logging (e.g., printing usage)

cmd/pg-schema-diff/apply_cmd.go
```diff
@@ -20,37 +20,51 @@ func buildApplyCmd() *cobra.Command {
...
+			toSchema:         toSchema,
+			tempDbConnConfig: connConfig,
+			planOptions:      planOptions,
+			logger:           logger,
+		})
 		if err != nil {
 			return err
 		} else if len(plan.Statements) == 0 {
-			fmt.Println("Schema matches expected. No plan generated")
+			cmd.Println("Schema matches expected. No plan generated")
```
bplunkett-stripe, line 62: Canonically, all logs should be done through the command. We were just doing this wrong previously. This is how cobra does it's own logging (e.g., printing usage)

internal/pgengine/engine.go
```diff
@@ -190,9 +193,9 @@ func (e *Engine) testIfInstanceServingTraffic() error {
 func (e *Engine) GetPostgresDatabaseConnOpts() ConnectionOptions {
 	result := make(map[ConnectionOption]string)
 	result[ConnectionOptionDatabase] = "postgres"
-	result["host"] = e.sockPath
-	result["user"] = e.superuser
-	result["port"] = strconv.Itoa(defaultPort)
+	result[ConnectionOptionHost] = e.sockPath
```
bplunkett-stripe, line 196: Exposed so tests can access them for the `--{from/to}-empty-dsn` parameters

cmd/pg-schema-diff/flags.go
```diff
@@ -7,29 +7,51 @@ import (
...
 
-type connFlags struct {
-	dsn string
+type connectionFlags struct {
+	// dsn is the connection string for the database.
+	dsn         string
+	dsnFlagName string
+
+	// isEmptyDsnUsingPq indicates to connect via DSN using the pq environment variables and defaults.
+	isEmptyDsnUsingPq         bool
```
bplunkett-stripe, line 18: [This was requested per this PR](https://github.com/stripe/pg-schema-diff/pull/54). This maintains support but makes it more explicit and flexible

cmd/pg-schema-diff/flags.go
```diff
@@ -7,29 +7,51 @@ import (
...
+	isEmptyDsnUsingPqFlagName string
 }
 
-func createConnFlags(cmd *cobra.Command) *connFlags {
-	flags := &connFlags{}
+func createConnectionFlags(cmd *cobra.Command, prefix string, additionalHelp string) *connectionFlags {
```
alexaub-stripe, line 22: Why is additionalHelp passed in? createConnectionFlags only appears to be used once, so it could be inlined.

Likewise with from- and to- I'm realizing.

alexaub-stripe: Ahh I see the plan cmd file was hidden for me

cmd/pg-schema-diff/plan_cmd.go
```diff
@@ -203,13 +259,56 @@ func timeoutModifierFlagVar(cmd *cobra.Command, p *[]string, timeoutType string,
 	cmd.Flags().StringArrayVarP(p, flagName, shorthand, nil, description)
 }
 
-func parsePlanConfig(p planFlags) (planConfig, error) {
-	schemaSourceFactory, err := parseSchemaSource(p.dbSchemaSourceFlags)
-	if err != nil {
-		return planConfig{}, err
+func parseSchemaSource(p schemaSourceFactoryFlags) (schemaSourceFactory, error) {
+	// Store result in a var instead of returning early to ensure only one option is set.
```
alexaub-stripe, line 263: Couldn't you just early return if not `len(p.schemaDirs) > 0  XOR p.connFlags.IsSet()`? Should remove a couple of the ssf nil check error blocks below.

alexaub-stripe (changes requested):
Just one comment worth looking at, but nothing blocking! Looks great!

cmd/pg-schema-diff/plan_cmd.go
bplunkett-stripe: I prefer this approach because it scales slightly better. If you put the condition at the top like you suggest then we are duplicating the condition, meaning they can become out of alignment.

You can imagine this gets more painful if we scale to 3 separate cases. Imagine a new schema source type that we use if `foobarParams != ""`. We would then need to add it in two separate places with the approach you suggest

alexaub-stripe: Fine to go with this, though I think you sacrifice some readability in order to avoid that duplication of the checks.

Half joking but I wonder if pushing ssf's onto a result slice and then waiting til the end to check if there isn't exactly one would be the best of both worlds. No need to do ssf != nil checks on each case.

alexaub-stripe (approved):

cmd/pg-schema-diff/plan_cmd.go
bplunkett-stripe: >Fine to go with this, though I think you sacrifice some readability in order to avoid that duplication of the checks.
Too true

>Half joking but I wonder if pushing ssf's onto a result slice and then waiting til the end to check if there isn't exactly one would be the best of both worlds. No need to do ssf != nil checks on each case.
I half agree 😎

Issue #148, opened 2024-07-20
Create Go script that can replace preceding tabs with spaces in SQL string literals


ammiranda:
Related to #67, this issue is a follow-up to the PR resolving #67 which will add a Go based script that can be run as needed to correct the whitespace for the string literals.

The desired feature set should include:

* Re-writes the acceptance tests to have tab-less multi-line strings
* Preferably has a--fix option, such that there is basically a validation/dry-run mode and a write mode
* Is called from the makefile lint and lint_fix targets (like go_lint, sql_lint, etc)

ammiranda:
@bplunkett-stripe Feel free to update the issue as needed.

bplunkett-stripe:
Looks right to me!

bplunkett-stripe:
You still planning on picking this up @ammiranda? Checking because I might take at a pass at it soon.

ammiranda:
@bplunkett-stripe I was planning on it, but I think I need a bit more direction on how to implement it. Like how to make it invocable from the command line via the Makefile and how to read the `--fix` flag for it. I'm guessing the flag could be read using the `flags` package but I'm unsure.

bplunkett-stripe:
Yes, you can use the same cobra/flags pkg, similar to the pg-schema-diff command. I would define the script under a file like `scripts/acceptance_test_sql_linter/main.go`. You could then use `go run` to call it from the makefile, passing the appropriate parameters.

PR #202, opened 2025-04-11
feat: #148 adding go script and makefile commands

ammiranda:
### Description

Adding a Go script that will do the whitespace tab to four spaces replacement for the acceptance test files.

### Motivation
Resolves issue #148 

### Testing

I ran the script both from the root of the directory and via the Makefile commands and observed they ran as expected.


ammiranda:
Hey @bplunkett-stripe I would love a review on this when you get a chance, thanks!

bplunkett-stripe (approved):
The script may need a few tweaks. Let's give it a try for now and see how it works.

PR #219, opened 2025-07-22
Accelerate tests

bplunkett-stripe:
### Description
Stop using a test suite in the acceptance tests. This enables us to leverage `t.Parallel`. Because tests ran in parallel, to maintain the same "randomly" generated outputs, I wired through a rand reader rather than relying on a global singleton. This makes tests about 50% faster

Before:
image

After:
image

### Motivation
Speed up acceptance tests.
### Testing
[//]: # Describe how you tested these changes
Ran locally


bplunkett-stripe:
Bypassing review since changes small in scope.

PR #233, opened 2025-08-24
Fix unit test, by adding a fixed restrict key for pgdump output comparison

velagent:
### Description
Fix unit test, by adding a fixed restrict key for pgdump output comparison

### Motivation
To get unit tests to pass 
Without this fix, cmd/pg-schema-diff/apply_cmd_test.go fails because the restrict key defaults to a random guid, and won't match when comparing two different pgdump outputs

### Testing
go test


bplunkett-stripe:
This is blocking other development, so I am forward fixing in a separate PR off of your branch. Your authorship on the initial contribution is still tracked, and you should be added to the list of contributors! Sorry about that!

PR #268, opened 2026-02-09
Account relation dependencies used by routines as composite types

joas8211:
### Description

Functions and procedures can use tables and views as composite types for parameter and return types. Those dependencies must be accounted for to avoid out of order errors in diffs.

Resolves #248

### Motivation

I'm building a project which uses table as function parameter type and couldn't get it working without these changes.

### Testing

Test cases are still missing which is why this PR is marked as draft. I'll add tests soon. I'd be happy to take feedback even before writing tests.

### DIFF
--- a/internal/queries/queries.sql
+++ b/internal/queries/queries.sql
@@ -266,7 +266,33 @@ SELECT
     pg_catalog.pg_get_function_identity_arguments(
         pg_proc.oid
     ) AS func_identity_arguments,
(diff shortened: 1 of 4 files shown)


joas8211:
The fix didn't actually work for me. It's complaining about circular dependencies. I have a big schema already so debugging it is a chore. I'd be happy if someone else took this inspiration from this to resolve the issue. I'll close this PR because I don't have time to finish it currently.

PR #270, opened 2026-03-05
Fix: recreate dependent views when altering column type           

feelixs:
### Description

In `buildViewDiff`, after checking for deleted columns, also check if any dependent column's type is being altered. If so, mark the view for recreation (drop before, recreate after).

### Motivation

When a table column's type is changed,  views that depend on it are not dropped first and recreated later, causing ALTER TABLE...ALTER COLUMN...SET DATA TYPE to fail:

`Error: 0A000: cannot alter type of a column used by a view or rule`

### Testing

Added three acceptance test cases in `view_cases_test.go`:
  - Single column type change with dependent view
  - Multiple column type changes with dependent view
  - Non-dependent column type change (view should NOT be recreated)

  Without the fix, the first two tests fail with the Postgres error above.


feelixs:
closes #280

Issue #284, opened 2026-05-15
Views that call functions are not ordered after those functions


da77a:
# Views that call functions are not ordered after those functions

## pg-schema-diff version

v1.0.5

## Problem

A view definition can call a function:

```sql
CREATE FUNCTION fn_compute(id integer) RETURNS TABLE(result text) ...;
CREATE VIEW v_computed AS SELECT * FROM fn_compute(42);
```

pg-schema-diff does not track this dependency. The view may be created before the function, causing:

```
ERROR: function fn_compute(integer) does not exist (SQLSTATE 42883)
```

## Root cause

The view vertex generator (`view_sql_generator.go`) only wires dependencies on tables (via `TableDependencies`). It has no awareness of function dependencies.

PostgreSQL does track view→function dependencies in `pg_depend` (via the view's rewrite rule), so the information is available:

```sql
SELECT pg_proc.proname, proc_ns.nspname,
       pg_catalog.pg_get_function_identity_arguments(pg_proc.oid)
FROM pg_catalog.pg_depend AS d
JOIN pg_catalog.pg_rewrite AS r ON d.objid = r.oid AND r.ev_class = <view_oid>
JOIN pg_catalog.pg_depend AS d2 ON r.oid = d2.objid
JOIN pg_catalog.pg_proc ON d2.refobjid = pg_proc.oid AND d2.refclassid = 'pg_proc'::REGCLASS
JOIN pg_catalog.pg_namespace AS proc_ns ON pg_proc.pronamespace = proc_ns.oid
WHERE d.refobjid = <view_oid> AND d2.deptype = 'n'
  AND proc_ns.nspname NOT IN ('pg_catalog', 'information_schema');
```

## Suggested fix

1. Add a `function_dependencies` column to `GetViews` (similar query to above)
2. Add `DependsOnFunctions []SchemaQualifiedName` to the `View` struct
3. In `view_sql_generator.go` `Add()`, wire: `mustRun(addVertexId).after(buildFunctionVertexId(f, diffTypeAddAlter))`

## Reproducer

```sql
CREATE FUNCTION public.fn_helper() RETURNS TABLE(x int) LANGUAGE sql AS $$ SELECT 1; $$;
CREATE VIEW public.v_uses_fn AS SELECT * FROM public.fn_helper();
```

Plan from empty → target: view is ordered before function → apply fails.

## Related

- #248 (function dependency ordering)

Issue #283, opened 2026-05-15
View dependency query produces false positives and misses view-to-view dependencies


da77a:
# View dependency query produces false positives and misses view-to-view dependencies

## pg-schema-diff version

v1.0.5

## Problem

The `GetViews` query in `internal/queries/queries.sql` has two issues in its `table_dependencies` subquery:

### 1. Missing view-to-view dependencies

The join filter `dep_c.relkind IN ('r', 'p')` excludes views (`'v'`). Views that depend on other views (e.g. `CREATE VIEW a AS SELECT * FROM other_view`) have no dependency edge, so they can be ordered before the view they reference.

**Fix:** Change to `dep_c.relkind IN ('r', 'p', 'v')` and add `dep_c.oid != c.oid` to exclude self-references (every view's rewrite rule has a dependency on itself).

### 2. False-positive dependencies from other views' rewrite rules

The join path:
```sql
FROM pg_catalog.pg_depend AS d
INNER JOIN pg_catalog.pg_rewrite AS r ON d.objid = r.oid
INNER JOIN pg_catalog.pg_depend AS d2 ON r.oid = d2.objid
...
WHERE d.refobjid = c.oid
```

This finds rewrite rules that reference the current view (`d.refobjid = c.oid`), then finds what those rules depend on. But OTHER views' rewrite rules also reference this view (if they SELECT from it). This causes view A to appear to depend on view B's dependencies, creating false cycles.

**Fix:** Add `AND r.ev_class = c.oid` to the `pg_rewrite` join to restrict to the view's own rewrite rule only.

## Reproducer

```sql
CREATE TABLE t1 (id int PRIMARY KEY, val text);
CREATE VIEW v1 AS SELECT * FROM t1;
CREATE VIEW v2 AS SELECT * FROM v1;  -- v2 depends on v1
```

Running `pg-schema-diff plan --from-dsn empty --to-dsn target`:
- Without fix: v2 may be ordered before v1 (missing dep), or cycle detected (false positive from other rules)
- With fix: correct order t1 → v1 → v2

## Related

- #248 (function composite type ordering — different issue but same dependency-tracking gap)

Issue #282, opened 2026-05-15
Function body table/view references not tracked for dependency ordering


da77a:
# Function body table/view references not tracked for dependency ordering

## pg-schema-diff version

v1.0.5

## Problem

Functions that reference tables or views in their body (via FROM/JOIN clauses, or %ROWTYPE/%TYPE declarations) have no dependency edge to those objects. This causes functions to be ordered before the tables/views they reference:

```
ERROR: relation "myschema.mytable" does not exist (SQLSTATE 42P01)
```

This affects:
- `LANGUAGE sql` functions — body is opaque text (`$function$...$function$`), PostgreSQL does not record table deps in `pg_depend`
- `LANGUAGE plpgsql` functions — same, plus `%ROWTYPE` and `type[]` declarations in DECLARE blocks are resolved at runtime, not tracked in `pg_depend`

The `HAS_UNTRACKABLE_DEPENDENCIES` hazard is emitted for plpgsql functions, but no ordering attempt is made even when the information is available in the function definition text.

## Patterns not tracked

1. **Schema-qualified FROM/JOIN:** `FROM schema.table_name` or `JOIN schema.view_name`
2. **%ROWTYPE declarations:** `DECLARE r schema.view_name%ROWTYPE`
3. **Array-of-composite declarations:** `DECLARE batch schema.view_name[]`

All of these are visible in `pg_get_functiondef()` output.

## Suggested fix

Something better than this - which I'm using as "good enough for now":

Supplementing  `pg_depend`-based detection with text-based parsing of the function definition. Regex patterns:

```
(?i)\b(\w+)\.(\w+)(?:%ROWTYPE|\[\])     -- schema.name%ROWTYPE or schema.name[]
(?i)(?:FROM|JOIN)\s+(\w+)\.(\w+)\b      -- FROM/JOIN schema.name
```

Matches are added as `TableDependencies` on the function, which the existing vertex generator already wires correctly.

## Reproducer

```sql
CREATE TABLE public.orders (id int PRIMARY KEY, total numeric);
CREATE VIEW public.order_summary AS SELECT id, total FROM public.orders WHERE total > 0;

CREATE FUNCTION public.process() RETURNS void LANGUAGE plpgsql AS $$
DECLARE r order_summary%ROWTYPE;
BEGIN
  FOR r IN SELECT * FROM order_summary LOOP NULL; END LOOP;
END; $$;
```

[...]

PR #286, opened 2026-05-16
Add function, view, and materialized view dependency ordering

da77a:
## Function and view dependency ordering

This PR adds dependency tracking for functions, views, and materialized views so that `pg-schema-diff` (mostly) correctly orders DDL statements when these objects reference each other.

### Motivation
Less cases where manual migration authoring is needed due to cascading dependencies issues involving objects other than tables.

This is the second attempt after some "working except when it doesn't" problems found using the first attempt.

### Approach
This attempt created failing test cases systematically for the unsupported/failing scenarios, then fixed (most of) them.

### What works
  **From-scratch creation** — given any schema containing interdependent functions, views, and materialized views, the generated plan creates objects in the correct order:

  - Functions that reference tables/views via `FROM`/`JOIN` (qualified and unqualified)
  - Functions that use `%ROWTYPE` or `[]` composite type declarations
  - Views that call functions
  - Views that depend on other views
  - Materialized views that depend on views, functions, or other materialized views
  - Cross-schema dependencies

  **Incremental migration** — when a table change forces a view to be recreated (column rename/delete), dependent views and materialized views are cascaded:

  - View A depends on View B, View B is recreated → View A is also recreated
  - Materialized view depends on a view being recreated → matview is also recreated
  - Materialized view depends on another matview being recreated → cascades correctly
  - Multi-level chains (A → B → C → table) propagate correctly

  **Identical-schema diffs** — diffing a schema against itself produces an empty plan, even when views depend on other views or call functions.

### Known gaps (with failing tests documenting each)

  | Gap | Reason |
  |-----|--------|
  | Comma-separated `FROM t1, t2, t3` — only first table detected | Regex cannot distinguish from column lists without a parser |
  | Function calling another function in body | pg_depend does not track fn→fn for SQL functions; body-text detection only covers table/view refs |
  | Function return type change requiring drop+create | `CREATE OR REPLACE` is always emitted, PG rejects if return type differs |

  ### Approach

  1. **pg_depend-based tracking** for structural dependencies (view→table columns, view→function, matview→function) via subqueries in  `GetViews`/`GetMaterializedViews`/`GetProcs`
  2. **Body-text regex** for dependencies pg_depend does not track (plpgsql `%ROWTYPE`, `[]`, and `FROM`/`JOIN` references in function bodies) — best-effort,  used for ordering only
  3. **Cascade propagation** after the initial diff pass — if a view/matview is being recreated, anything that depends on it is also marked for recreation
  4. **Cycle prevention** — reverse ordering edges (function must be altered before its old table dep) are only emitted for dependencies being *removed*, not persisting ones

  ### Testing

Acceptance test cases covering all the above scenarios, including the documented gaps (marked with `expectedPlanErrorContains`).

### DIFF
--- a/cmd/pg-schema-diff/plan_cmd.go
+++ b/cmd/pg-schema-diff/plan_cmd.go
@@ -117,6 +117,7 @@ type (
 		dataPackNewTables     bool
 		disablePlanValidation bool
 		noConcurrentIndexOps  bool
+		skipPrivileges        bool
 
 		statementTimeoutModifiers []string
 		lockTimeoutModifiers      []string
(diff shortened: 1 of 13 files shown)


da77a:
Fixes #248 #282 #283 #284

Related:
PR https://github.com/stripe/pg-schema-diff/pull/268 (partial fix abandoned due to cycles)
and https://github.com/stripe/pg-schema-diff/pull/285 (first attempt at a better fix)

PR #295, opened 2026-07-02
fix: escape enum labels in generated SQL to prevent SQL injection

jtayal-stripe:
## Summary

Enum labels are interpolated into generated DDL statements using raw `fmt.Sprintf("'%s'", val)`. A label containing a single quote breaks out of the string literal, allowing arbitrary SQL to be injected into the generated migration.

This is a **second-order SQL injection**: a user with `CREATE` privilege on any schema plants a malicious enum label, which is then executed with the **plan runner's privileges** (typically a superuser/DBA) when `pg-schema-diff` generates migration SQL. Because Postgres allows `COPY ... TO PROGRAM`, this escalates to **remote code execution** on the database host.

## Vulnerable sinks (`pkg/diff/enum_sql_generator.go`)

- `CREATE TYPE ... AS ENUM (...)` — enum values in `Add`
- `ALTER TYPE ... ADD VALUE '...'` — new value in `Alter`
- `... BEFORE '...'` — the ordering anchor in `Alter`

## Fix

- Add `EscapeLiteral()` to `internal/schema` alongside the existing `EscapeIdentifier()`. It doubles single quotes per the SQL standard and strips null bytes, following the same strategy as `pgx`'s `QuoteString`.
- Route all three sinks through `schema.EscapeLiteral(...)` instead of raw interpolation.

## Tests

Three layers, all passing:

- **Unit** (`internal/schema/escape_test.go`): `EscapeLiteral` across normal values, single/multiple quotes, empty strings, backslashes, null bytes, and injection payloads.
- **Unit** (`pkg/diff/enum_sql_generator_test.go`): generator output for `Add`, `ADD VALUE`, and `ADD VALUE ... BEFORE` produces correctly escaped DDL.
- **Acceptance** (`internal/migration_acceptance_tests/enum_cases_test.go`): end-to-end cases (apply plan against real Postgres, then compare `pg_dump`) for CREATE, ADD VALUE, ADD VALUE ... BEFORE, and an injection-payload regression (`x'); DROP TABLE foo; --`). If escaping regresses, the payload would drop the table and the dump comparison would fail.

### Test output

```
--- PASS: TestEnumTestCases/create_enum_with_single-quote_labels
--- PASS: TestEnumTestCases/add_value_with_single-quote_label
--- PASS: TestEnumTestCases/add_value_before_a_single-quote_label
--- PASS: TestEnumTestCases/enum_label_with_sql_injection_payload
ok  github.com/stripe/pg-schema-diff/internal/migration_acceptance_tests

--- PASS: TestEscapeLiteral (8 subtests)
ok  github.com/stripe/pg-schema-diff/internal/schema

--- PASS: TestEnumSQLGenerator_Add_EscapesLabels / _Alter_EscapesLabels / _Alter_EscapesBEFORE
ok  github.com/stripe/pg-schema-diff/pkg/diff
```

## How to test

```bash
docker build -t pg-schema-diff-test-runner -f ./build/Dockerfile.test --build-arg PG_VERSION=14 .
docker run --rm pg-schema-diff-test-runner -v -race ./internal/migration_acceptance_tests/ -run 'TestEnumTestCases' -timeout 10m
```

### DIFF
--- a/internal/migration_acceptance_tests/enum_cases_test.go
+++ b/internal/migration_acceptance_tests/enum_cases_test.go
@@ -118,6 +118,88 @@ var enumAcceptanceTestCases = []acceptanceTestCase{
 		// as a validation error. In the future, we can identify this in the actual plan generation stage.
 		expectedPlanErrorContains: errValidatingPlan.Error(),
 	},
+	{
+		// Exercises the CREATE TYPE ... AS ENUM path with labels containing single
+		// quotes. Verifies the generated DDL is correctly escaped, executes against
+		// Postgres, and stores the labels verbatim.
+		name: "create enum with single-quote labels",
+		oldSchemaDDL: []string{
+			`
+            CREATE TABLE foo();
+			`,
+		},
+		newSchemaDDL: []string{
+			`
+            CREATE TYPE quote_enum AS ENUM ('O''Brien', 'it''s', 'plain');
+            CREATE TABLE foo(
+                val quote_enum DEFAULT 'plain'
+            );
+			`,
+		},
+	},
+	{
+		// Exercises the ALTER TYPE ... ADD VALUE path (value appended at the end)
+		// with a label containing a single quote.
+		name: "add value with single-quote label",
+		oldSchemaDDL: []string{
+			`
+            CREATE TYPE quote_enum AS ENUM ('a', 'b');
+            CREATE TABLE foo(
+                val quote_enum
+            );
+			`,
+		},
+		newSchemaDDL: []string{
+			`
+            CREATE TYPE quote_enum AS ENUM ('a', 'b', 'won''t');
+            CREATE TABLE foo(
+                val quote_enum
+            );
+			`,
+		},
+	},
+	{
+		// Exercises the ALTER TYPE ... ADD VALUE ... BEFORE path. The new value is
+		// inserted in the middle, so a BEFORE clause is emitted referencing a
+		// following label that contains a single quote.
+		name: "add value before a single-quote label",
+		oldSchemaDDL: []string{
+			`
+            CREATE TYPE quote_enum AS ENUM ('a', 'it''s');
+            CREATE TABLE foo(
+                val quote_enum
+            );
+			`,
+		},
+		newSchemaDDL: []string{
+			`
+            CREATE TYPE quote_enum AS ENUM ('a', 'mid', 'it''s');
+            CREATE TABLE foo(
+                val quote_enum
+            );
+			`,
+		},
+	},
+	{
+		// Regression test for the enum-label SQL injection: a label crafted as an
+		// injection payload must be stored inertly, not executed. If escaping ever
+		// regresses, applying the generated plan would attempt to DROP the table and
+		// the pg_dump comparison would fail.
+		name: "enum label with sql injection payload",
+		oldSchemaDDL: []string{
+			`
+            CREATE TABLE foo();
+			`,
+		},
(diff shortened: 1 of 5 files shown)


alexaub-stripe (changes requested):
This looks good to me, and I think these tests are correct and cover the bad behavior, but could you attach some evidence that they fail when the escaping isn't implemented?

jtayal-stripe:
Good call @alexaub-stripe. I reverted the escaping in `enum_sql_generator.go` (restoring the raw `fmt.Sprintf("'%s'", val)` in all three sinks) while keeping the new tests, and re-ran them. Both the unit and acceptance layers fail as expected.

### Unit tests (`pkg/diff`) — without the fix

```
--- FAIL: TestEnumSQLGenerator_Add_EscapesLabels/label_with_single_quote
    expected: CREATE TYPE "public"."color" AS ENUM ('it''s', 'fine')
    actual  : CREATE TYPE "public"."color" AS ENUM ('it's', 'fine')
--- FAIL: TestEnumSQLGenerator_Add_EscapesLabels/SQL_injection_attempt
    expected: CREATE TYPE "public"."color" AS ENUM ('x''); DROP TABLE foo; --')
    actual  : CREATE TYPE "public"."color" AS ENUM ('x'); DROP TABLE foo; --')
--- FAIL: TestEnumSQLGenerator_Alter_EscapesLabels
    expected: ALTER TYPE "public"."status" ADD VALUE 'it''s complicated'
    actual  : ALTER TYPE "public"."status" ADD VALUE 'it's complicated'
--- FAIL: TestEnumSQLGenerator_Alter_EscapesBEFORE
    expected: ALTER TYPE "public"."status" ADD VALUE 'new''val' BEFORE 'it''s complicated'
    actual  : ALTER TYPE "public"."status" ADD VALUE 'new'val' BEFORE 'it's complicated'
FAIL	github.com/stripe/pg-schema-diff/pkg/diff
```

### Acceptance tests (real Postgres) — without the fix

```
--- FAIL: TestEnumTestCases/create_enum_with_single-quote_labels
    CREATE TYPE "public"."quote_enum" AS ENUM ('O'Brien', 'it's', 'plain'): ERROR: syntax error at or near "Brien" (SQLSTATE 42601)
--- FAIL: TestEnumTestCases/add_value_with_single-quote_label
    ALTER TYPE "public"."quote_enum" ADD VALUE 'won't': ERROR: syntax error at or near "t" (SQLSTATE 42601)
--- FAIL: TestEnumTestCases/add_value_before_a_single-quote_label
    CREATE TYPE "public"."quote_enum" AS ENUM ('a', 'it's'): ERROR: syntax error at or near "s" (SQLSTATE 42601)
--- FAIL: TestEnumTestCases/enum_label_with_sql_injection_payload
    DDL generated: CREATE TYPE "public"."evil" AS ENUM ('x'); DROP TABLE foo; --')
    ERROR: relation "public.foo" does not exist (SQLSTATE 42P01)
FAIL	github.com/stripe/pg-schema-diff/internal/migration_acceptance_tests
```

The injection case is the clearest demonstration: without escaping, the generated statement becomes `CREATE TYPE ... AS ENUM ('x'); DROP TABLE foo; --')`, and Postgres then reports `relation "public.foo" does not exist` — i.e. the injected `DROP TABLE foo` actually executed against the database. With the fix in place, all of these pass. I've restored the fix; no code change from this exercise.

alexaub-stripe (approved):
That `relation "public.foo" does not exist` is perfect, thank you!

PR #298, opened 2026-07-20
fix: qualify pg_catalog.to_timestamp in bigint-to-timestamp migrations

rajsuvariya-stripe:
## Summary

When migrating a `bigint` column to `timestamp without time zone`, pg-schema-diff emits a `to_timestamp()` call in the `ALTER COLUMN ... USING` clause. An unqualified call resolves via `search_path`, so a user with `CREATE` on a schema can plant shadow functions that run during apply instead of the built-ins (CVE-2018-1058). On a default Postgres install, `search_path` is `"$user", public, pg_catalog` — `public` is already searched before `pg_catalog`.

**Fix:** emit a fully qualified expression:

```sql
pg_catalog.to_timestamp(
  col::pg_catalog.float8 OPERATOR(pg_catalog./) 1000.0::pg_catalog.float8
)
```

The shadow acceptance case plants `public.to_timestamp(double precision)` because the fixed emission passes **float8** to `to_timestamp` (not bigint). That matches the overload that would win if `pg_catalog.` qualification were removed while keeping the float8 casts. The original pre-fix emission `to_timestamp(col / 1000)` uses bigint division and is covered by the "without fix" evidence below.

## Evidence of vulnerability without fix

Reverted `generateTypeTransformationStatement` to pre-fix emission (`to_timestamp(col / 1000)`) and ran the shadow acceptance test locally (PostgreSQL 14.15 via pgengine, default `search_path`).

```
ERROR:  pwnd
STATEMENT:  ... using to_timestamp("some_time_col" / 1000);

    acceptance_test.go:182: ERROR: pwnd (SQLSTATE P0001)
--- FAIL: TestColumnTestCases/Change_BIGINT_to_TIMESTAMP_(ignores_shadowed_to_timestamp_in_USING)
```

Unqualified `to_timestamp()` in the emitted `USING` clause invokes the planted shadow during apply.

The validate case asserts qualified DDL in `expectedPlanDDL`; without the fix the generator emits `using to_timestamp("some_time_col" / 1000)`.

## Evidence of tests passing with fix

Restored the fix and re-ran both acceptance tests locally (PostgreSQL 14.23 via pgengine, `pg_dump` 14.23 with `--restrict-key`, default `search_path`):

```bash
export PATH="/opt/homebrew/opt/postgresql@14/bin:$PATH"
go test ./internal/migration_acceptance_tests/... \
  -run 'TestColumnTestCases/Change_BIGINT_to_TIMESTAMP' -count=1 -v
```

```
=== RUN   TestColumnTestCases/Change_BIGINT_to_TIMESTAMP_(validate_conversion_and_ANALYZE)
=== RUN   TestColumnTestCases/Change_BIGINT_to_TIMESTAMP_(ignores_shadowed_to_timestamp_in_USING)
--- PASS: TestColumnTestCases (0.00s)
    --- PASS: TestColumnTestCases/Change_BIGINT_to_TIMESTAMP_(validate_conversion_and_ANALYZE) (1.92s)
    --- PASS: TestColumnTestCases/Change_BIGINT_to_TIMESTAMP_(ignores_shadowed_to_timestamp_in_USING) (1.92s)
PASS
ok  	github.com/stripe/pg-schema-diff/internal/migration_acceptance_tests	4.943s
```

Both cases pass the full acceptance harness: **plan → apply → pg_dump schema compare**. Apply succeeds with no `ERROR: pwnd` in Postgres logs (the shadow is not invoked).

CI `run_tests` (PG 14, 15, 16, 17) passed on `490295f` and `794276a`.

## Tests

- **Change BIGINT to TIMESTAMP (validate conversion and ANALYZE)** — `expectedPlanDDL` for qualified emission + ANALYZE
- **Change BIGINT to TIMESTAMP (ignores shadowed to_timestamp in USING)** — planted `public.to_timestamp(double precision)` + row data; apply succeeds only with the fix

```bash
go test ./internal/migration_acceptance_tests/... \
  -run 'TestColumnTestCases/Change_BIGINT_to_TIMESTAMP' -count=1
```

## Test plan

- [x] Standard bigint→timestamp case asserts plan DDL + ANALYZE
- [x] Shadow case verifies apply succeeds against planted `public.to_timestamp(double precision)` on default search_path
- [x] Verified shadow case fails with `pwnd` when fix is reverted to `to_timestamp(col / 1000)`
- [x] Local acceptance tests pass end-to-end on PG 14.23 (plan/apply/pgdump)
- [x] CI green on PG 14–17

### DIFF
--- a/internal/migration_acceptance_tests/column_cases_test.go
+++ b/internal/migration_acceptance_tests/column_cases_test.go
@@ -327,8 +327,44 @@ var columnAcceptanceTestCases = []acceptanceTestCase{
 			`,
 		},
 		expectedPlanDDL: []string{
-			"ALTER TABLE \"public\".\"Foobar\" ALTER COLUMN \"some_time_col\" SET DATA TYPE timestamp without time zone using to_timestamp(\"some_time_col\" / 1000)",
+			"ALTER TABLE \"public\".\"Foobar\" ALTER COLUMN \"some_time_col\" SET DATA TYPE timestamp without time zone using pg_catalog.to_timestamp(\"some_time_col\"::pg_catalog.float8 OPERATOR(pg_catalog./) 1000.0::pg_catalog.float8)",
 			"ANALYZE \"public\".\"Foobar\" (\"some_time_col\")",
+		},
+		expectedHazardTypes: []diff.MigrationHazardType{
+			diff.MigrationHazardTypeAcquiresAccessExclusiveLock,
+			diff.MigrationHazardTypeImpactsDatabasePerformance,
+		},
+	},
+	{
+		name: "Change BIGINT to TIMESTAMP (ignores shadowed to_timestamp in USING)",
+		oldSchemaDDL: []string{
+			`
+            CREATE FUNCTION public.to_timestamp(double precision)
+            RETURNS timestamp without time zone LANGUAGE plpgsql AS $$
+            BEGIN
+              RAISE EXCEPTION 'pwnd';
+            END $$;
+
+            CREATE TABLE "Foobar"(
+                id INT PRIMARY KEY,
+                some_time_col BIGINT NOT NULL
+            );
+            INSERT INTO "Foobar" VALUES (1, 1700000000000);
+			`,
+		},
+		newSchemaDDL: []string{
+			`
+            CREATE FUNCTION public.to_timestamp(double precision)
+            RETURNS timestamp without time zone LANGUAGE plpgsql AS $$
+            BEGIN
+              RAISE EXCEPTION 'pwnd';
+            END $$;
+
+            CREATE TABLE "Foobar"(
+                id INT PRIMARY KEY,
+                some_time_col TIMESTAMP NOT NULL
+            );
+			`,
 		},
 		expectedHazardTypes: []diff.MigrationHazardType{
 			diff.MigrationHazardTypeAcquiresAccessExclusiveLock,
--- a/pkg/diff/sql_generator.go
+++ b/pkg/diff/sql_generator.go
@@ -1399,7 +1399,7 @@ func (csg *columnSQLVertexGenerator) generateTypeTransformationStatement(
 	if strings.EqualFold(oldType, "bigint") &&
 		strings.EqualFold(newType, "timestamp without time zone") {
 		return Statement{
-			DDL: fmt.Sprintf("%s SET DATA TYPE %s using to_timestamp(%s / 1000)",
+			DDL: fmt.Sprintf("%s SET DATA TYPE %s using pg_catalog.to_timestamp(%s::pg_catalog.float8 OPERATOR(pg_catalog./) 1000.0::pg_catalog.float8)",
 				csg.alterColumnPrefix(col),
 				newType,
 				schema.EscapeIdentifier(col.Name),


cmd/pg-schema-diff/apply_cmd.go
```diff
@@ -149,8 +149,8 @@ func runPlan(ctx context.Context, cmd *cobra.Command, connConfig *pgx.ConnConfig
 		if _, err := conn.ExecContext(ctx, fmt.Sprintf("SET SESSION lock_timeout = %d", stmt.Timeout.Milliseconds())); err != nil {
 			return fmt.Errorf("setting lock timeout: %w", err)
 		}
-		if _, err := conn.ExecContext(ctx, stmt.ToSQL()); err != nil {
-			return fmt.Errorf("executing migration statement. the database maybe be in a dirty state: %s: %w", stmt.DDL, err)
+		if err := diff.ExecuteStatement(ctx, conn, stmt); err != nil {
+			return fmt.Errorf("the database maybe be in a dirty state: %w", err)
```
alexaub-stripe: Why remove the "executing migration statement:" context and the stmt.ddl here?

internal/migration_acceptance_tests/acceptance_test.go
```diff
@@ -243,11 +243,19 @@ func applyDDL(db *pgengine.DB, ddl []string) error {
...
+
+	ctx := context.Background()
 	for _, stmt := range plan.Statements {
-		ddl = append(ddl, stmt.ToSQL())
+		if err := diff.ExecuteStatement(ctx, conn, stmt); err != nil {
+			return err
+		}
 	}
-	return applyDDL(db, ddl)
+	return nil
```
alexaub-stripe: What's the point of this refactor? It's reimplementing the applyDDL implementation right above it

internal/migration_acceptance_tests/search_path_cases_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stripe/pg-schema-diff/internal/pgengine"
+	"github.com/stripe/pg-schema-diff/pkg/diff"
+	"github.com/stripe/pg-schema-diff/pkg/log"
+	"github.com/stripe/pg-schema-diff/pkg/sqldb"
+	"github.com/stripe/pg-schema-diff/pkg/tempdb"
+)
+
+// Regression for CVE-2018-1058 / SAT-28189: a shadow public.to_timestamp(bigint)
```
alexaub-stripe: SAT-28189 is an internal reference and shouldn't be included here.

pkg/diff/sql_generator.go
```diff
@@ -1399,13 +1399,14 @@ func (csg *columnSQLVertexGenerator) generateTypeTransformationStatement(
...
+			DDL: fmt.Sprintf("%s SET DATA TYPE %s using pg_catalog.to_timestamp((%s / 1000.0)::pg_catalog.float8)",
 				csg.alterColumnPrefix(col),
 				newType,
 				schema.EscapeIdentifier(col.Name),
 			),
-			Timeout:     statementTimeoutDefault,
-			LockTimeout: lockTimeoutDefault,
+			Timeout:                 statementTimeoutDefault,
+			LockTimeout:             lockTimeoutDefault,
+			pinSearchPathBeforeExec: true,
```
alexaub-stripe: I'm not sure we need this pinSearchPathBeforeExec feature. Including `pg_catalog.` above resolves the CVE as the search path won't be used to find `pg_catalog.to_timestamp` at all. 

If we wanted search path pinning as some kind of broad safety check, it should be implemented for all sql applied, not just this one statement that no longer needs it because we're correctly prefixing it. I'm not sure the value is there though -- we should just always prefix.

Let me know if I'm misunderstanding the CVE though -- the postgres guide [here](https://wiki.postgresql.org/wiki/A_Guide_to_CVE-2018-1058%3A_Protect_Your_Search_Path) was useful for understanding it.

alexaub-stripe (changes requested):
Fixing pg_catalog.to_timestamp makes sense, but I don't think the pinSearchPathBeforeExec feature should be implemented this way.

internal/schema/cast_type.go
```diff
@@ -0,0 +1,61 @@
+package schema
+
+import "strings"
+
+// formatTypeBaseToPgCatalogTypname maps pg_catalog.format_type base names to pg_catalog
+// typnames suitable for CAST(... AS type). format_type returns human-readable names
+// (e.g. "integer") but CAST requires typnames (e.g. int4) when schema-qualified.
+var formatTypeBaseToPgCatalogTypname = map[string]string{
```
jtayal-stripe: The map covers the most common numeric and temporal types, but several other types that `format_type()` returns unqualified are missing:

> - `text` → `text`
> - `bytea` → `bytea`
> - `json` → `json`
> - `jsonb` → `jsonb`
> - `uuid` → `uuid`
> - `date` → `date`
> - `interval` → `interval`

For these, the fallback emits `"pg_catalog"."text"` (double-quoted) instead of `pg_catalog.text`. Both execute correctly, but the quoted form looks unusual in generated DDL and could confuse users reading migration output. Worth populating the map more completely - a quick way to enumerate candidates is `SELECT typname FROM pg_type WHERE typnamespace = 'pg_catalog'::regnamespace AND typtype = 'b'`.

internal/schema/cast_type.go
```diff
@@ -0,0 +1,61 @@
...
+	if isSchemaQualifiedTypeName(typeName) {
+		return typeName
+	}
+
+	base, typmod := splitTypeNameAndTypmod(typeName)
+	if typname, ok := formatTypeBaseToPgCatalogTypname[strings.ToLower(base)]; ok {
+		return "pg_catalog." + typname + typmod
+	}
+
+	return `"pg_catalog".` + EscapeIdentifier(base) + typmod
```
jtayal-stripe:
```go
 return `"pg_catalog".` + EscapeIdentifier(base) + typmod
 ```

This assumes every unrecognized unqualified type lives in `pg_catalog`. If a user-defined type somehow arrives here without schema qualification (e.g., `myenum`), it would generate `"pg_catalog"."myenum"` which fails at apply time. I believe pg-schema-diff's internal representation always schema-qualifies user types, so this shouldn't happen in practice, but a short comment documenting that assumption would help future readers.

**Review by jtayal-stripe — DISMISSED** [review]
LGTM !!

File: internal/migration_acceptance_tests/column_cases_test.go [hunk 7]
```diff
@@ -737,7 +737,7 @@ var columnAcceptanceTestCases = []acceptanceTestCase{
 			"ALTER TABLE \"public\".\"foobar\" ADD CONSTRAINT \"pgschemadiff_tmpnn_EBESExQVRheYGRobHB0eHw\" CHECK(\"foobar\" IS NOT NULL) NOT VALID",
 			"ALTER TABLE \"public\".\"foobar\" VALIDATE CONSTRAINT \"pgschemadiff_tmpnn_EBESExQVRheYGRobHB0eHw\"",
 			"ALTER TABLE \"public\".\"foobar\" ALTER COLUMN \"foobar\" SET NOT NULL",
-			"ALTER TABLE \"public\".\"foobar\" ALTER COLUMN \"foobar\" SET DATA TYPE integer using \"foobar\"::integer",
+			"ALTER TABLE \"public\".\"foobar\" ALTER COLUMN \"foobar\" SET DATA TYPE integer using CAST(\"foobar\" AS pg_catalog.int4)",
```
alexaub-stripe: We're scoping the cast, but we're not scoping the `SET DATA TYPE integer`? Doesn't that leave the same shadowing opportunity?

internal/schema/cast_type.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// QualifyTypeForCast returns a type name suitable for CAST(... AS type) in generated
+// migration SQL. Unqualified built-in types from pg_catalog.format_type are qualified
+// with pg_catalog to prevent search_path shadowing of cast targets.
+func QualifyTypeForCast(typeName string) string {
+	typeName = strings.TrimSpace(typeName)
+	if isSchemaQualifiedTypeName(typeName) {
+		return typeName
+	}
```
alexaub-stripe: pg_catalog.format_type doesn't qualify the type name if it's in the search path (ie `pg_catalog.format_type('public.mytype'::regtype, NULL);` returns `mytype` not `public.mytype`). So I don't think this will catch intentional user types whenever they're in the default search path?

alexaub-stripe (changes requested):
The to_timestamp change looks good: we're injecting an intentional pg_catalog conversion call in the bigint->timestamp case and we definitely want to use the system function. 

The type scoping changes are less clear to me: 
* Users should be able to use custom types in their schemas. I think the cast_type behavior breaks that as-is (see comment).
* The behavior in cast_type seems pretty coupled to postgres internals, and seems like it could break in surprising ways as postgres types evolve. If we wanted to do that properly, we should probably get a fully qualified type name at read time (pg-schema-diff/internal/queries/queries.sql#L66) and propagate that through the column diffs instead of assuming it's a pg_catalog type at alter time. 
* I don't think typename's suffer the same issue as functions described in CVE-2018-1058: for functions a better matching public function will shadow the pg_catalog entry with the default search path. But for typenames that's not the case since there's no better match behavior: unless the user intentionally puts pg_catalog in their search path after public, postgres will check pg_catalog first and take that type over anything in public.

Based on that last point, is there any reason to scope typenames in these alter statements at all? Maybe we can just make the to_timestamp change and leave the `USING` behavior as-is?

rajsuvariya-stripe:
@alexaub-stripe Per your review — reverted the cast qualification changes in 07b3322. Scope is now only the `pg_catalog.to_timestamp` fix for the bigint→timestamp migration.

- Removed `QualifyTypeForCast` / `CAST(... AS pg_catalog.*)`; generic transforms back to `col::type`
- Removed `TestGenericCastMigrationUsesQualifiedTargetType`
- Unqualified `SET DATA TYPE integer` left as-is (typename shadowing != function shadowing under CVE-2018-1058)

Ready for another look when you have a moment.

internal/migration_acceptance_tests/search_path_shadow_cases_test.go
```diff
@@ -0,0 +1,162 @@
+package migration_acceptance_tests
```
alexaub-stripe: Can this be restructured into the pattern used by the rest of the acceptance tests via the [acceptanceTestCase approach](https://github.com/stripe/pg-schema-diff/blob/main/internal/migration_acceptance_tests/acceptance_test.go#L32-L65)?

migration acceptance tests are intended to all use the same consistent full e2e plan & apply flow to verify user-facing correctness and avoid breaking when internal implementation details change. They're not unit tests and shouldnt be doing their own setup and behavior injection like this file does. 

TBH I think the to_timestamp shadow case could just be included in column_cases_test.sql. A way we can specifically verify the shadow behavior is avoided in an acceptance test is something like (pseudocode):
```
#planOpts should have an insert statement to ensure to_timestamp gets run ondata
planOpts: insert statement at index 0: insert into foo values (some_bigint_timestamp)

# old ddl with shadowing to_timestamp:
create function public.to_timestamp(input): RAISE EXCEPTION 'pwnd';
create table foo with bigint column;

# new ddl with bigint->timestamp conversion
create function public.to_timestamp(input): RAISE EXCEPTION 'pwnd';
create table foo with timestamp column
```
If public.to_timestamp can shadow, it'll raise an error when called on the inserted row and fail the case. Otherwise the correct pg_catalog.to_timestamp is used and the case passes.

alexaub-stripe (changes requested):
Looks good! But one more ask on testing

alexaub-stripe (approved):
Looks great! Thank you for all the cycling

PR #303, opened 2026-07-30
Fix multiline expression index handling

Pybsama:
### Description

`pg_get_indexdef` can preserve newlines inside expression index definitions. The existing anchored rewrite uses `.*`, which does not match newlines in Go, so converting a valid definition to `CREATE INDEX CONCURRENTLY` fails. The same conversion is also used while validating plans with concurrent index operations disabled.

This change enables dot-all matching only for the existing regex, preserving its anchors, case sensitivity, and exact `CREATE [UNIQUE] INDEX ` prefix. It adds regression coverage for:

- the direct statement conversion;
- adding a multiline expression index through the default concurrent path;
- validating a migration with an existing multiline expression index under `WithNoConcurrentIndexOps()`.

### Testing

- `GOTOOLCHAIN=go1.25.5 go test -race ./... -timeout 30m`
- focused conversion test, 100 repetitions
- both PostgreSQL 17.10 acceptance scenarios, 10 repetitions
- `golangci-lint v2.7.0 run`
- `go vet ./...`
- `sqlfluff 3.3.0 lint`
- `gofmt`, `go mod tidy -diff`, and `git diff --check`

The commit is GitHub Verified.

Fixes #253

### DIFF
--- a/internal/migration_acceptance_tests/index_cases_test.go
+++ b/internal/migration_acceptance_tests/index_cases_test.go
@@ -133,6 +133,33 @@ var indexAcceptanceTestCases = []acceptanceTestCase{
 		},
 	},
 	{
+		name: "Add a multiline expression index",
+		oldSchemaDDL: []string{
+			`
+            CREATE TABLE index_test (
+                flag BOOLEAN,
+                value1 TEXT,
+                value2 TEXT
+            );
+			`,
+		},
+		newSchemaDDL: []string{
+			`
+            CREATE TABLE index_test (
+                flag BOOLEAN,
+                value1 TEXT,
+                value2 TEXT
+            );
+            CREATE INDEX ix_test ON index_test ((
+                CASE WHEN flag THEN value1 ELSE value2 END
+            ));
+			`,
+		},
+		expectedHazardTypes: []diff.MigrationHazardType{
+			diff.MigrationHazardTypeIndexBuild,
+		},
+	},
+	{
 		name: "Add a unique index",
 		oldSchemaDDL: []string{
 			`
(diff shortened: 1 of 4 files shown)


Pybsama:
Revalidated this change against current `main` at `9ada4710c28718312557c84cca1be4b2557eed54`. The one commit since the PR base only changes materialized-view handling; the PR applies without conflicts, and the exact current-main merge tree is `6e3c23613b66d86b32aae0f226f2a7fefadaab50`.

Current-main validation:

- restoring the old single-line regex makes all three regressions fail independently: direct conversion, default concurrent index creation, and `WithNoConcurrentIndexOps()` plan validation
- restoring `(?s)` makes all three pass
- Go 1.25.5: full `go test -race ./... -timeout 30m` passed; the acceptance package completed in 145.747s
- focused conversion: 100/100 repetitions
- both PostgreSQL 17.10 acceptance scenarios: 10/10 repetitions
- golangci-lint 2.7.0 with Go 1.25.5: 0 issues; `go vet ./...` passed
- sqlfluff 3.3.0: passed; gofmt, `go mod tidy -diff`, and `git diff --check`: clean
- the multiline-string dry-run still names only the two unchanged baseline files (`policy_cases_test.go` and `privilege_cases_test.go`); neither new acceptance-test file is reported

The public head remains the original GitHub-verified commit because GitHub reports the PR as mergeable and the current-main change does not overlap the patch.

`license/cla` is still pending and the CLA assistant reports it as not signed. I am therefore keeping the PR in Draft until the contributor completes that personal legal step; no CLA action or recheck was performed here.

